PR #3, opened 2024-09-03
Implement CI

AndreyMlashkin:


crypto3/crypto3.nix
```diff
@@ -33,6 +31,7 @@ in stdenv.mkDerivation {
       (if enableDebug then "-DCMAKE_BUILD_TYPE=Debug" else "-DCMAKE_BUILD_TYPE=Release")
       (if enableDebug then "-DCMAKE_CXX_FLAGS=-ggdb" else "")
```
akokoshn, line 32: let's keep CXX flags in CMake (like in zkEVMFramework)

.github/workflows/linux-nix-check.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Clean up after previous checkout
+        run: chmod +w -R ${GITHUB_WORKSPACE}; rm -rf ${GITHUB_WORKSPACE}/*;
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Run checks
+        run: nix build -L .?#checks.x86_64-linux.all-gcc
```
akokoshn, line 21: why only build without check?

AndreyMlashkin: There are two chains to build now - gcc and clang.
It would take too long to build all. I would prefer to start 2 different jobs, so the CI will run faster

crypto3/crypto3.nix
AndreyMlashkin: Done

akokoshn (approved):

PR #6, opened 2024-09-09
Directly include transpiler

AndreyMlashkin:


transpiler/.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "cmake/modules"]
```
akokoshn, line 1: remove

transpiler/.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,18 @@
+name: Handle pull request event
```
akokoshn, line 1: in my understanding, should be only one `.gihub/workflow` for all repository.
could you please move logic to the root and check another folders (crypto3, ....)

transpiler/.gitmodules
AndreyMlashkin: done

transpiler/.github/workflows/pull_request.yml
AndreyMlashkin: removed, logic is already moved

akokoshn (approved):

ETatuzova (approved):

PR #15, opened 2024-09-16
add assigner app

AndreyMlashkin:


flake.nix
```diff
@@ -216,13 +216,17 @@
           default = all-gcc;
         };
         apps = {
+          assigner = {
+            type = "app";
+            program = "${self.packages.${system}.evm-assigner}/bin/assigner/assigner";
```
akokoshn, line 221: why `.evm-assigner`? executable is part of `zkevm-framework`

flake.nix
```diff
@@ -216,13 +216,17 @@
...
+          };
           single-threaded = {
             type = "app";
             program = "${self.packages.${system}.proof-producer}/bin/proof-producer-single-threaded";
           };
           default = {
```
akokoshn, line 227: default option is required?

AndreyMlashkin: yes, a proof-producer script required it. I can change the script and remove the default, if you think, it's better

flake.nix
AndreyMlashkin: this command works for me:
```
nix run  #assigner
```

I see, that the target is declared in zkevm-framework/tests/libs/assigner_runner/CMakeLists.txt, but it is deployed to the assigner folder...

AndreyMlashkin: I have corrected the package name now. I have no idea, why it was working before

flake.nix
akokoshn: Yes, thanks!. it's a little bit strange when required build some  application i any case

AndreyMlashkin: fixed

akokoshn (approved):

PR #17, opened 2024-09-16
Initial commit with custom big intger classes.

tigran-at-nil:
Please review initial codes for our big integer library.


akokoshn:
@martun , @ETatuzova , PR still actual?

PR #32, opened 2024-09-19
add precompiled headers

AndreyMlashkin:


crypto3/.gitignore
```diff
@@ -250,7 +250,6 @@ TSWLatexianTemp*
 
 # Precompiled Headers
 *.gch
-*.pch
```
akokoshn, line 253: Let's make single common .gitignore in the root of placeholder

crypto3/libs/algebra/include/nil/crypto3/algebra/algebra.pch
```diff
@@ -0,0 +1,13 @@
+#include <nil/crypto3/algebra/curves/pallas.hpp>
```
akokoshn, line 1: auto generated `*.pch `should be ignored, right?

crypto3/.gitignore
AndreyMlashkin: ok, I will do a general cmake-cleanup together with export statements unification

crypto3/libs/algebra/include/nil/crypto3/algebra/algebra.pch
AndreyMlashkin: This whole PR is yet highly experimental

PR #45, opened 2024-09-25
Fix find msb undefined behaviour

AndreyMlashkin:


vo-nil (approved):

tigran-at-nil (approved):
Reviewed. Should work fine for all current cases.

PR #60, opened 2024-10-03
Fix flake.nix

ioxid:
Fixes `nix flake show`.


ioxid:
Dropping as I found the fix in another pull request.

PR #61, opened 2024-10-03
Split transpiler tests, add KZG tests

ioxid:


transpiler/test/evm.cpp
```diff
@@ -0,0 +1,1632 @@
...
+
+        return true;
+    }
+
+    plonk_table_description<field_type> desc;
+};
+
+BOOST_AUTO_TEST_SUITE(placeholder_circuit2_kzg_v2)
+
+using TestFixtures = boost::mpl::list<
```
ETatuzova, line 1568: Please, add nil::crypto3::algebra::curves::alt_bn128_254::base_field_type curve here.

ETatuzova (changes requested):

transpiler/test/evm.cpp
```diff
@@ -0,0 +1,1632 @@
...
+    const plonk_table_description<typename PlaceholderParams::field_type> &table_description,
+    std::string folder_name
+) {
+    std::filesystem::create_directory(folder_name);
+    test_placeholder_proof<Endianness, placeholder_proof<typename PlaceholderParams::field_type, PlaceholderParams>>(
+        proof,
+        commitment_scheme.get_commitment_params(),
+        folder_name + "/proof.bin"
+    );
+    test_placeholder_partial_proof<
```
ETatuzova, line 386: Remove it. Not used here

transpiler/test/evm.cpp
ioxid: Oh, right, it's the one with built-in support in EVM. Done.

ioxid:
I found that verifier printer parameters in tests are probably incorrect, because the comments mention wrong parameters. This is probably a result of some refactoring. Since I don't know what these parameters mean I don't really know what should be there. But it can break in unexpected ways.
Screenshot_20241007_105457

(notice that 15 is passed for an optional boolean parameter, and two other parameter names don't match)

ETatuzova (approved):

ETatuzova (approved):

ETatuzova (approved):

ETatuzova (approved):

PR #88, opened 2024-10-15
86 todos in algebra

vo-nil:
This PR addresses TODOs in algebra code, including, but not limited  to:
    - handling errors from marshalling;
    - removal of extended_integral_type from fields;
    - final_exponentiation in pairings can fail when dealing with zero;
    - updated type_traits system to handle types properly;
    - fixed multiexp implementation and added tests for it;
    - added documentation references to constants origins;
    - removed old r1cs, ppzksnark code;
    - excluded evm-assigner and zkevm-framework from builds as they are
      deprecated and subject for removal;


crypto3/libs/algebra/include/nil/crypto3/algebra/curves/detail/jubjub/params.hpp
```diff
@@ -115,12 +103,11 @@ namespace nil {
                         template<typename Coordinates>
                         using group_type = jubjub_types::g1_type<forms::montgomery, Coordinates>;
 
-                        // TODO: check correctness of the base point coordinates
                         constexpr static const std::array<typename field_type::value_type, 2> one_fill = {
                             typename field_type::value_type(
                                 0x52a47af6ec47deb77d663b6a45b148d1ccdaa4e2299ecfbd5504c409b3ea62c0_cppui_modular255),
                             typename field_type::value_type(
-                                0x399b020832f6a499ba9d5334ca932dc9faaeea860e9a49a8f8854f55f3b676d4_cppui_modular254)};
```
martun, line 123: Can you write a comment describing from where we got this values?

crypto3/libs/algebra/include/nil/crypto3/algebra/fields/detail/element/fp12_2over3over2.hpp
```diff
@@ -143,11 +143,6 @@ namespace nil {
                             return *this;
                         }
 
-                        element_fp12_2over3over2 sqrt() const {
```
martun, line 146: I see we have sqrt for fp2. Doesn't that algorithm apply here? Are we sure we don't need sqrt for this extension fields?

crypto3/libs/algebra/include/nil/crypto3/algebra/fields/maxprime.hpp
```diff
@@ -49,8 +49,6 @@ namespace nil {
                     constexpr static const std::size_t modulus_bits = policy_type::modulus_bits;
                     typedef typename policy_type::integral_type integral_type;
```
martun, line 51: Doesn't look like we need this file.

crypto3/libs/algebra/include/nil/crypto3/algebra/multiexp/policies.hpp
```diff
@@ -220,7 +220,6 @@ namespace nil {
                         typedef typename std::iterator_traits<InputBaseIterator>::value_type base_value_type;
                         typedef typename std::iterator_traits<InputFieldIterator>::value_type field_value_type;
 
-                        // TODO(martun): check that we did not break this, since integral_type is now a fixed size type.
                         typedef typename field_value_type::integral_type non_fixed_precision_number_type;
```
martun, line 223: Can you make sure we have tests that cover this function? Also please rename non_fixed_precision_number_type to something else, like integral_type, because it is actually a fixed precision number.

crypto3/libs/algebra/test/fields_static.cpp
```diff
@@ -134,10 +110,6 @@ BOOST_AUTO_TEST_CASE(field_operation_test_bls12_381_fr) {
                                       0x53530b5d9374f4f45c1b2267998b8cd7d1bf6d19a90564fb898ed8970544ea28_cppui_modular255};
     constexpr const_set_t constants1 = {811706348};
 
-    // TODO: the reason of the error "function parameter 'elements' with unknown value cannot be used in a constant
```
martun, line 137: It's a bit strange this did not compile. Can you please re-try? Maybe we need to remove the const reference for elements1?

crypto3/libs/algebra/include/nil/crypto3/algebra/fields/detail/element/fp12_2over3over2.hpp
vo-nil: We need sqrt only for fields that can be used as base fields for curves.
For pairing-friendly curves these are:
Fp2 for MNT4::G2
Fp3 for MNT6::G2
Fp2 for BLS12::G2

Sqrt is specifically used for marshalling a compressed form of a point - we store only X (or Y) coordinate, then solve the equation, like this: $y^2= x^3+bx+c$ to recover Y coordinate.

Fields for GT groups - Fp4 (GT for MNT4), Fp6 (GT for MNT6)  and Fp12  (GT for BLS12) do not serve as base fields for elliptic curve elements and thus do not require sqrt function. In most places it was empty with TODO.

crypto3/libs/algebra/include/nil/crypto3/algebra/fields/maxprime.hpp
vo-nil: File removed

crypto3/libs/algebra/test/fields_static.cpp
vo-nil: I have found a trick that worked: we need to have values `static` and pass them as template parameters by value to make that function work.
This introduced some caveat, I can't figure out how to properly check presence of `.sqrt()` method.

crypto3/libs/algebra/include/nil/crypto3/algebra/multiexp/policies.hpp
vo-nil: Added test for multiexp functions

martun (approved):

PR #92, opened 2024-10-16
Add function fresh_subcontext.

martun:


crypto3/libs/blueprint/include/nil/blueprint/bbf/generic.hpp
```diff
@@ -172,44 +176,96 @@ namespace nil {
...
+                        desc.constant_columns, std::vector<bool>(desc.usable_rows_amount));
+                }
+
+                bool is_allocated(std::size_t col, std::size_t row, column_type t) {
+                    if (col >= log[t].size() || row >= log[t][col].size()) {
+                        std::stringstream error;
+                        error << "Invalid value col = " << col << ", row = " << row 
+                            << " when checking if a " << t << " cell is allocated. We have "
+                            << log[t].size() << " columns each having "
+                            << log[t][col].size() << " rows.";
```
ayashunsky, line 200: This will go out of range if we had col >= log[t].size(), so we should rather do smth like log[t][0].size(), but in fact we are not guaranteed to have columns of a specific type at all, so maybe we should just shorten this error report.

crypto3/libs/blueprint/include/nil/blueprint/bbf/generic.hpp
```diff
@@ -172,44 +176,96 @@ namespace nil {
...
+                    return log[t][col][row];
+                }
+
+                void mark_allocated(std::size_t col, std::size_t row, column_type t) {
+                    if (col >= log[t].size() || row >= log[t][col].size()) {
+                        std::stringstream error;
+                        error << "Invalid value col = " << col << ", row = " << row 
+                            << " when marking a " << t << " cell allocated. We have "
+                            << log[t].size() << " columns each having "
+                            << log[t][col].size() << " rows.";
```
ayashunsky, line 212: Here same as above, [col] may be out of range.

crypto3/libs/blueprint/include/nil/blueprint/bbf/bbf_wrapper.hpp
```diff
@@ -216,7 +216,14 @@ namespace nil {
                 using component_type = plonk_bbf_wrapper<BlueprintFieldType>;
                 using var = typename component_type::var;
 
-                context_type ct = context_type(assignment,8,start_row_index); // max_rows = 8
+                // get_description() function returns the actual number of rows used. In order to allow more rows to be used we
+                // must add the max_rows to the returned sizes.
+                std::size_t max_rows = 8;
+                auto desc = assignment.get_description();
+                desc.usable_rows_amount += max_rows;
```
ayashunsky, line 223: at assignment-genereation stage we don't do anything like that, so we end up with two allocation_logs of different sizes, which is probably not so good.

ayashunsky (commented):
I'd stay we still need some more changes. See comments above.

crypto3/libs/blueprint/include/nil/blueprint/bbf/bbf_wrapper.hpp
martun: Changed this. Now the caller needs to know, that the description provided is actually the used part.

crypto3/libs/blueprint/include/nil/blueprint/bbf/generic.hpp
martun: Changed.

crypto3/libs/blueprint/include/nil/blueprint/bbf/generic.hpp
martun: Done.

ayashunsky (approved):

PR #98, opened 2024-10-18
Multiprecision refactoring

ioxid:


crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/big_uint_impl.hpp
```diff
@@ -0,0 +1,1674 @@
...
+    template<std::size_t Bits>
+    std::ostream& operator<<(std::ostream& os, const big_uint<Bits>& value) {
+        os << value.str(os.flags());
+        return os;
+    }
+
+    // Common ops
+
+    template<std::size_t Bits>
+    constexpr std::size_t msb(const big_uint<Bits>& a) {
```
martun, line 1648: Why do we need this helper? Maybe just change all the calls of msb(a) to a.msb() and delete this, and other helpers?

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/integer_ops_powm.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include "nil/crypto3/multiprecision/big_uint_impl.hpp"
+#include "nil/crypto3/multiprecision/ops/powm.hpp"
+
+namespace nil::crypto3::multiprecision {
+    template<typename T1, typename T2, typename T3,
+             std::enable_if_t<std::numeric_limits<std::decay_t<T1>>::is_integer &&
+                                  std::numeric_limits<std::decay_t<T2>>::is_integer &&
+                                  std::is_integral_v<T3>,
+                              int> = 0>
+    constexpr T3 powm(T1 &&b, T2 &&e, T3 m) {
```
martun, line 15: Maybe not in this PR, but we can remove all these functions, and make them members of some classes.

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/modular/ops/pow.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace nil::crypto3::multiprecision {
+    template<typename big_mod_t, typename T,
+             std::enable_if_t<detail::is_big_mod_v<big_mod_t> &&
+                                  std::numeric_limits<T>::is_integer,
+                              int> = 0>
+    constexpr big_mod_t pow(big_mod_t b, T e) {
+        if constexpr (std::is_signed_v<T>) {
+            if (e < 0) {
+                b = inverse_extended_euclidean_algorithm(b);
+                // TODO(ioxid): this won't work for e == min()
```
martun, line 31: Create a task for yourself to not forget and fix this later, if it actually will work incorrectly for min(). We can leave optimization TODOs forever, but the cases where it can return incorrect result must be addressed soon.

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/ops/import_export.hpp
```diff
@@ -0,0 +1,210 @@
...
+            using size_type = typename std::make_unsigned<difference_type>::type;
+
+            if (!chunk_size) {
+                chunk_size = std::numeric_limits<value_type>::digits;
+            }
+
+            size_type limbs = std::distance(i, j);
+            size_type bits = limbs * chunk_size;
+
+            // TODO(ioxid): this breaks marshalling tests
```
martun, line 127: I think we need another check here, if I understand correctly. Very often we have 255 bit numbers, or 381 bit numbers. In this case marshalling may give you 4 limbs of 64 bits and 6 limbs of 64 bits respectively, since 255 and 381 are not divisible by 64. What you should check here is that there extra bits are ALL ZERO! 

And better have a throw here, not an assert, since it may happen in production, if someone intentionally gives us a wrong marshalled file as input.

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/ops/import_export.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+            // This will remove the upper bits using upper_limb_mask.
+            newval.normalize();
+
+            result = std::move(newval);
+            return result;
+        }
+
+        template<std::size_t Bits, typename T>
+        inline big_uint<Bits>& import_bits_fast(big_uint<Bits>& result, T* i, T* j,
```
martun, line 148: Maybe move this, and other functions that must be used only from this file to an unnamed namespace, such that people won't call them from the outside. Kind of make them "private".

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/ops/powm.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include "nil/crypto3/multiprecision/big_uint_impl.hpp"
+#include "nil/crypto3/multiprecision/modular/big_mod_impl.hpp"
+#include "nil/crypto3/multiprecision/modular/ops/pow.hpp"
+
+namespace nil::crypto3::multiprecision {
+    template<typename T1, std::size_t Bits, typename T2,
+             std::enable_if_t<std::numeric_limits<std::decay_t<T1>>::is_integer &&
+                                  std::numeric_limits<std::decay_t<T2>>::is_integer,
+                              int> = 0>
+    constexpr big_uint<Bits> powm(T1 &&b, T2 &&e, const big_uint<Bits> &m) {
```
martun, line 29: Maybe we can give up on powm, in a later commit? We needed it for integration with boost, now we're done...

crypto3/libs/transpiler/test/circuit1_lpc/gate_argument.sol
```diff
@@ -0,0 +1,74 @@
+
+// SPDX-License-Identifier: Apache-2.0.
+//---------------------------------------------------------------------------//
+// Copyright (c) 2023 Generated by ZKLLVM-transpiler
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
martun, line 6: These files were not supposed to be in this PR. I think you unintentionally added some .sol files here.

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/big_uint_impl.hpp
ioxid: Yeah. So there is some code that should work for both big_uint and built-in integers. This code uses these free functions.

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/big_uint_impl.hpp
```diff
@@ -0,0 +1,1674 @@
...
+#include "nil/crypto3/multiprecision/detail/endian.hpp"
+#include "nil/crypto3/multiprecision/detail/intel_intrinsics.hpp"
+#include "nil/crypto3/multiprecision/storage.hpp"
+
+namespace nil::crypto3::multiprecision {
+    template<std::size_t Bits>
+    class big_uint;
+
+    namespace detail {
+        constexpr bool is_valid_hex_digit(char c) {
```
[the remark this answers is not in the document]
ioxid: Constexpr on functions implies inline (https://en.cppreference.com/w/cpp/language/constexpr). Unnamed namespace would be worse because the parent namespace would be polluted.

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/integer_ops_powm.hpp
ioxid: This should work for built-in types as well

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/ops/powm.hpp
ioxid: It's used by miller-rabin. I could probably rewrite it to used big_mod_rt, but this is a generally useful function too

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/ops/import_export.hpp
ioxid: In headers this wont actually work since they will be available for [qualified name lookup](https://en.cppreference.com/w/cpp/language/qualified_lookup) in parent namespace. I think that detail namespace better protects them from usage.

crypto3/libs/transpiler/test/circuit1_lpc/gate_argument.sol
ioxid: Yeah, sure, I will remove them

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/big_int.hpp
```diff
@@ -0,0 +1,399 @@
...
+                return other.m_abs.compare(this->m_abs);
+            }
+            return this->m_abs.compare(other.m_abs);
+        }
+
+        // Arithmetic operations
+
+        // Addition/subtraction
+
+        static constexpr big_int<Bits + 1> add(const big_int& a, const big_int& b) noexcept {
```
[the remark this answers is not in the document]
ioxid: I have plans to refactor this, but we first need to discuss what we do with overflow here.

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/ops/import_export.hpp
ioxid: Added this to the comment. I think I can do this after initial merge.

martun (approved):

martun (approved):

PR #130, opened 2024-11-01
Fix : Nix installation command by correctly using the -L flag in curl…

Budalebah:
… command

correction of  the usage of the -L flag in the Nix installation command, ensuring that curl handles redirects properly. The -L flag is now correctly placed with a space separating it from the URL to avoid argument misinterpretation. This change resolves an error where curl was not following redirects as intended, improving the reliability of the installation command for all users

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -16,7 +16,7 @@ Supported by [=nil; Foundation](https://nil.foundation)
 Install nix using the following command:
 
 ```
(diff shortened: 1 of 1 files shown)


martun (approved):

martun (changes requested):
This command looks ok to me, as is, no need to change.

Budalebah:
I tried it many times but got error with L , can you check again

PR #158, opened 2024-11-26
Implement benchmark report

AndreyMlashkin:


crypto3/benchmarks/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ set(BENCHMARK_NAMES
 
     "multiprecision/modular_adaptor_fixed"
 
-    "zk/lpc"
+    #"zk/lpc"
```
akokoshn, line 63: why disabled?

PR #162, opened 2024-11-28
fix error around CMAKE_BUILD_TYPE

AndreyMlashkin:


AndreyMlashkin:
@vo-nil , please confirm, that it fixes your problem

PR #168, opened 2024-12-02
unit test for copy circuit

oclaw:


oclaw:
probably will be merged in #172

PR #178, opened 2024-12-05
Split trace files

makxenov:


proof-producer/libs/assigner/include/nil/proof-generator/assigner/trace_parser.hpp
```diff
@@ -25,6 +25,11 @@ namespace nil {
             std::vector<blueprint::bbf::rw_operation> storage_ops;
         };
 
+        const char BYTECODE_EXTENSION[] = ".bc";
```
akokoshn, line 28: constants can be reused from https://github.com/NilFoundation/placeholder/blob/bdde215653dcf3bd964a2b3a94f2b9918aed03e3/proof-producer/libs/preset/include/nil/proof-generator/preset/preset.hpp#L23
May be move to common header: `limits.hpp` => `constants.hpp`

proof-producer/libs/assigner/include/nil/proof-generator/assigner/trace_parser.hpp
```diff
@@ -36,20 +41,62 @@ namespace nil {
...
+
+                executionproofs::RWTraces pb_traces;
+                if (!pb_traces.ParseFromIstream(&file)) {
                     return std::nullopt;
                 }
+                return pb_traces;
+            }
 
-                executionproofs::ExecutionTraces pb_traces;
+            [[nodiscard]] std::optional<executionproofs::ZKEVMTraces> read_zkevm_traces_from_file(const boost::filesystem::path& filename) {
```
akokoshn, line 64: Can we use single template function?

proof-producer/libs/assigner/include/nil/proof-generator/assigner/trace_parser.hpp
makxenov: These constants are different entities. In preset we have circuit names, and I introduce extension names for trace files. It is technically possible to unify them, but I think it would be more confusing.

akokoshn (approved):

oclaw (approved):

PR #183, opened 2024-12-06
proof-producer: rw & zkevm sat check fixes

oclaw:
Following changes:
- Added sort of rw_operations_vector before passing it into assignment + refactored copypaste in copy and zkevm circuits
- Removed some old comments
- Fixed build
- Add dummy impl for not implemented opcodes in order to enable checks for zkevm circuit
- Added traces with single call of increment and parametrized tests with them
- Enabled single tx check for rw and bytecode circuits

Now disabled checks are
- zkevm both single and multiple trace (cause is to be investigated)
- bytecode multiple trace (cause is to be investigated)
- rw multiple trace (most probably bug in circuit)


crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/dummy.hpp
```diff
@@ -0,0 +1,35 @@
...
+                virtual void fill_context(
+                    typename generic_component<FieldType, GenerationStage::ASSIGNMENT>::context_type &context,
+                    const opcode_input_type<FieldType, GenerationStage::ASSIGNMENT> &current_state
+                ) override  {
+                }
+                virtual void fill_context(
+                    typename generic_component<FieldType, GenerationStage::CONSTRAINTS>::context_type &context,
+                    const opcode_input_type<FieldType, GenerationStage::CONSTRAINTS> &current_state
+                ) override  {
+                    std::cout << "DUMMY CONTSTRAINT, REPLACE ME WITH SOME ACTUAL IMPL" << std::endl;
```
akokoshn, line 27: may be dummy constrain for utilize 1 row as reported in `rows_amount`

proof-producer/libs/assigner/include/nil/proof-generator/assigner/rw.hpp
```diff
@@ -38,6 +38,12 @@ namespace nil {
             for (const auto& storage_op : rw_operations->storage_ops) {
                 input.push_back(storage_op);
             }
+
+            using nil::blueprint::bbf::rw_operation;
```
akokoshn, line 42: Let's move sort to `deserialize_rw_traces_from_file`, sorted rw operation used by another circuits, like zkevm

oclaw: zkevm and other circuits do not require operations to be sorted, and I am not sure if it breaks something or not
Will check

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/dummy.hpp
oclaw: They are actually needed to skip row from zkevm table without breaking the checks, so seems OK to me

akokoshn: ok

proof-producer/libs/assigner/include/nil/proof-generator/assigner/rw.hpp
oclaw: Actually nothing broke, got rid of struct RWOperations and now do the sort inside trace parser

makxenov (approved):

akokoshn (approved):

PR #200, opened 2024-12-12
Optimize compilation

AndreyMlashkin:


.github/workflows/clang-linux-nix-check.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           nix build -L .?#checks.x86_64-linux.crypto3-clang
         env:
           NIX_CONFIG: |
-            cores = 4
+            cores = 16
```
martun, line 25: I see you were able to increase the number of cores. Did the memory usage decline after pre-compiled headers, and that allowed you to use more cores?

AndreyMlashkin: Yes, exactly!

crypto3/libs/marshalling/zk/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ target_include_directories(${CMAKE_WORKSPACE_NAME}_${CURRENT_PROJECT_NAME} INTER
 
 target_link_libraries(${CMAKE_WORKSPACE_NAME}_${CURRENT_PROJECT_NAME} INTERFACE
                       Boost::container
+                      Boost::random
```
akokoshn, line 29: may be make sense link `Boost` librarie only to executable (proof-producer and tests), not for header only libraries?
But we can do it in another PR

akokoshn (approved):

.github/workflows/gcc-linux-nix-check.yml
```diff
@@ -22,17 +22,13 @@ jobs:
           nix build -L .?#checks.x86_64-linux.crypto3-gcc
         env:
           NIX_CONFIG: |
-            cores = 4
-
+            cores = 12
```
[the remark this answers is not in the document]
AndreyMlashkin: Gcc is more RAM consuming, it was failing with cores = 16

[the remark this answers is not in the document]
AndreyMlashkin: I will give a try with 16 now

crypto3/libs/marshalling/zk/CMakeLists.txt
akokoshn: i mean, no need to link boost libraries with header only targets

proof-producer/libs/output_artifacts/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ endif()
 target_link_libraries(proof_generatorOutputArtifacts
                       PUBLIC
                       crypto3::common
-                      Boost::log
```
akokoshn, line 18: why log was removed?, as far i see it still used: https://github.com/NilFoundation/placeholder/blob/aaa4d94750a8ca3fe98f10ed6dd13c1d123055ea/proof-producer/libs/output_artifacts/include/nil/proof-generator/output_artifacts/assignment_table_writer.hpp#L20

AndreyMlashkin: Fixed

akokoshn (approved):

crypto3/libs/marshalling/zk/CMakeLists.txt
martun: But this way we keep track of which library uses which boost library. Doesn't help the compilation much, but probably does not hurt as well.

martun (approved):

PR #224, opened 2024-12-23
Add deb bundler for proof-producer package

makxenov:
- Fix protobuf package import to fetch absl dependencies automatically
- Enable building proof-producer binaries as statically linked
- Add bundler to make deb package
- Add package version as a revCount

### DIFF
--- a/.github/workflows/deb-package-proof-producer-bundler.yaml
+++ b/.github/workflows/deb-package-proof-producer-bundler.yaml
@@ -1,0 +1,23 @@
+name: Check Linux bundlers
+
+on:
+  workflow_call:
+
+jobs:
+  build-deb-package:
+    name: "Build proof-producer deb package"
+    runs-on: [self-hosted, Linux, X64, aws_autoscaling]
+    steps:
+      # https://github.com/actions/checkout/issues/1552
+      - name: Clean up after previous checkout
+        run: chmod +w -R ${GITHUB_WORKSPACE}; rm -rf ${GITHUB_WORKSPACE}/*;
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Build proof-producer deb package
+        run: |
+          nix bundle --bundler . .#proof-producer
+          ls -l ./deb-package-proof-producer/proof-producer_*_amd64.deb
--- a/flake.nix
+++ b/flake.nix
@@ -16,10 +16,22 @@
   outputs = { self, nixpkgs, flake-utils, nix-3rdparty, ... }:
     (flake-utils.lib.eachDefaultSystem (system:
       let
+        revCount = self.revCount or self.dirtyRevCount or 1;
+        version = "0.0.1-${toString revCount}";
         pkgs = import nixpkgs {
           inherit system;
           overlays = [ nix-3rdparty.overlays.${system}.default ];
         };
+
+        # For proof-producer, our main target is statically linked binaries,
+        # so we should pass static libraries as build inputs
+        staticOverlay = final: prev: {
+          boost = prev.pkgsStatic.boost.override{ enableShared = false;};
+          protobuf = (prev.pkgsStatic.protobuf.override { enableShared = false;});
+          gtest = (prev.pkgsStatic.gtest.override { static = true;});
+        };
+        staticPkgs = pkgs.extend staticOverlay;
+
       in rec {
         packages = rec {
           crypto3 = (pkgs.callPackage ./crypto3.nix {
@@ -80,19 +92,19 @@
             runTests = false;
           });
 
-          proof-producer = (pkgs.callPackage ./proof-producer.nix {
+          proof-producer = (staticPkgs.callPackage ./proof-producer.nix {
             runTests = false;
             enableDebug = false;
           });
-          proof-producer-tests = (pkgs.callPackage ./proof-producer.nix {
+          proof-producer-tests = (staticPkgs.callPackage ./proof-producer.nix {
             runTests = true;
             enableDebug = false;
           });
-          proof-producer-debug-tests = (pkgs.callPackage ./proof-producer.nix {
+          proof-producer-debug-tests = (staticPkgs.callPackage ./proof-producer.nix {
             enableDebug = true;
             runTests = true;
           });
-          proof-producer-clang-debug = (pkgs.callPackage ./proof-producer.nix {
+          proof-producer-clang-debug = (staticPkgs.callPackage ./proof-producer.nix {
             stdenv = pkgs.llvmPackages_19.stdenv;
             enableDebug = true;
             runTests = false;
@@ -102,6 +114,7 @@
           });
 
           develop = (pkgs.callPackage ./proof-producer.nix {
+            staticBuild = false;
             enableDebug = true;
             runTests = true;
             sanitize = true;
@@ -112,6 +125,7 @@
           });
 
           develop-clang = (pkgs.callPackage ./proof-producer.nix {
+            staticBuild = false;
             stdenv = pkgs.llvmPackages_19.stdenv;
             enableDebug = true;
             runTests = true;
@@ -187,16 +201,17 @@
             benchmarkTests = true;
           });
 
-          proof-producer-gcc = (pkgs.callPackage ./proof-producer.nix {
+          proof-producer-gcc = (staticPkgs.callPackage ./proof-producer.nix {
             runTests = true;
             enableDebug = false;
           });
-          proof-producer-clang = (pkgs.callPackage ./proof-producer.nix {
+          proof-producer-clang = (staticPkgs.callPackage ./proof-producer.nix {
             stdenv = pkgs.llvmPackages_19.stdenv;
             runTests = true;
             enableDebug = false;
           });
           proof-producer-clang-sanitize = (pkgs.callPackage ./proof-producer.nix {
+            staticBuild = false;
             stdenv = pkgs.llvmPackages_19.stdenv;
             runTests = true;
             enableDebug = false;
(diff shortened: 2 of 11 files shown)


proof-producer/libs/assigner/CMakeLists.txt
```diff
@@ -43,11 +41,9 @@ target_link_libraries(proof_generatorAssigner
                         PUBLIC
                         proof_generatorPreset
                         crypto3::common
+                        protobuf::libprotobuf
```
oclaw, line 44: god I love cmake conventions

.github/workflows/deb-package-proof-producer-bundler.yaml
```diff
@@ -0,0 +1,23 @@
...
+  build-deb-package:
+    name: "Build proof-producer deb package"
+    runs-on: [self-hosted, Linux, X64, aws_autoscaling]
+    steps:
+      # https://github.com/actions/checkout/issues/1552
+      - name: Clean up after previous checkout
+        run: chmod +w -R ${GITHUB_WORKSPACE}; rm -rf ${GITHUB_WORKSPACE}/*;
+
+      - name: Checkout
+        uses: actions/checkout@v3
```
x-mass, line 16:
```suggestion
        uses: actions/checkout@v4
```

flake.nix
```diff
@@ -97,6 +99,13 @@
             enableDebug = true;
             runTests = false;
           });
+          proof-producer-static-build = (pkgs.callPackage ./proof-producer.nix {
```
oclaw, line 102: other flakes do not have `-build` suffix, imo better omit it here too

flake.nix
```diff
@@ -227,5 +236,30 @@
...
+                cp -r *.deb $out
+              '';
+            };
+          default = deb;
+        };
       }));
 }
+
+# To make deb package with proof-producer:
+# nix bundle --bundler . .#proof-producer-static-build
```
oclaw, line 265: let's add empty line to make github happy :)

proof-producer/CMakeLists.txt
```diff
@@ -20,6 +20,16 @@ include(CMConfig)
 include(CMDeploy)
 include(CMSetupVersion)
 
+option(PROOF_PRODUCER_STATIC_BINARIES "Link proof-producer binaries statically" OFF)
+
+if(PROOF_PRODUCER_STATIC_BINARIES)
+    set(BUILD_SHARED_LIBS OFF)
+    set(Boost_USE_STATIC_LIBS ON)
+else()
+    set(BUILD_SHARED_LIBS OFF)  # TODO: fix issue with precompiled headers to build shared libraries
```
oclaw, line 29: Is it possible to refer to the issue somehow (or at least briefly describe it)?

proof-producer/bin/proof-producer/CMakeLists.txt
```diff
@@ -80,6 +80,10 @@ function(setup_proof_generator_target)
         target_compile_options(${ARG_TARGET_NAME} PRIVATE "-fconstexpr-ops-limit=4294967295")
     endif ()
 
+    if(PROOF_PRODUCER_STATIC_BINARIES)
+        target_link_options(${ARG_TARGET_NAME} PRIVATE -static -static-libgcc -static-libstdc++)
```
oclaw, line 84: Is there any reason for using `target_link_options` instead of `target_link_libraries`?

proof-producer/CMakeLists.txt
makxenov: I was trying to revive build with shared libraries (they were abandoned recently), but got this error:
```
       > cc1plus: error: /build/source/build/proof-producer/libs/output_artifacts/CMakeFiles/proof_generatorOutputArtifacts.dir/cmake_pch.hxx.gch: not used because `proof_generatorOutputArtifacts_EXPORTS' not defined [-Werror=invalid-pch]
       > cc1plus: all warnings being treated as errors
       > ninja: build stopped: subcommand failed.
```
To be honest, I still don't know what went wrong, and since this is not a subject of this PR, I decided to fix it later

proof-producer/bin/proof-producer/CMakeLists.txt
makxenov: Is there an option in `target_link_libraries` to force a linker use `-static` flag?

oclaw: https://cmake.org/cmake/help/latest/command/target_link_libraries.html
```
Added in version 3.13: [LINK_OPTIONS](https://cmake.org/cmake/help/latest/prop_tgt/LINK_OPTIONS.html#prop_tgt:LINK_OPTIONS) target property and [target_link_options()](https://cmake.org/cmake/help/latest/command/target_link_options.html#command:target_link_options) command. For earlier versions of CMake, use [LINK_FLAGS](https://cmake.org/cmake/help/latest/prop_tgt/LINK_FLAGS.html#prop_tgt:LINK_FLAGS) property instead.
```

I think `set_target_properties(${ARG_TARGET_NAME} LINK_OPTIONS ...)` + `target_link_libraries` is a bit more transparent approach

proof-producer/libs/assigner/CMakeLists.txt
```diff
@@ -25,13 +23,13 @@ foreach(PROTO_FILE ${PROTO_FILES})
     list(APPEND GENERATED_HEADERS ${PROTO_HDR})
 endforeach()
 
-add_library(proof_generatorAssigner STATIC
+add_library(proof_generatorAssigner
             ${PROTO_SRC}
```
AndreyMlashkin, line 27: Why?

proof-producer/libs/output_artifacts/CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
 add_library(proof_generatorOutputArtifacts 
-    STATIC
     src/output_artifacts.cpp
```
AndreyMlashkin, line 2: why?

flake.nix
```diff
@@ -97,6 +99,13 @@
             enableDebug = true;
             runTests = false;
           });
+          proof-producer-static-build = (pkgs.callPackage ./proof-producer.nix {
+            runTests = false;
+            enableDebug = false;
+            staticBuild = true;
+            boost = pkgs.pkgsStatic.boost.override{ enableShared = false;};
+            protobuf = (pkgs.pkgsStatic.protobuf.override { enableShared = false;});
```
AndreyMlashkin, line 107: Maybe easier would be to override boost and protobuf conditionally depending on staticBuild flag inside the derivation?
You can also set PROOF_PRODUCER_STATIC_BINARIES, BUILD_SHARED_LIBS and Boost_USE_STATIC_LIBS flags conditionally

proof-producer/bin/proof-producer/CMakeLists.txt
makxenov: AFAIS `target_link_libraries` doesn't allow me to set options conditionally (depending on `PROOF_PRODUCER_STATIC_BINARIES`). And using temporary variable or `set_target_properties` seems less clear to me then `target_link_options` that is intended for this usecase.

oclaw: target link options documentation actually is not clear about adding libraries though this call (it just says `any options can be passed, but there target_link_libraries call which is precisely for libraries`)

I do not think it is a big deal since linkage is not so complicated here, but cmake tends to provide some really unexpected side effects in such cases, so I am a bit concerned :)

anyways, up to you

proof-producer/libs/assigner/CMakeLists.txt
makxenov: Now it's possible to build either static or dynamic libraries depending on `BUILD_SHARED_LIBS` flag. But it's still static for all the derivations, due to the issue I described above

flake.nix
akokoshn: May be just use static build allays for all derivations?
I guess we have no special cases where required dynamic linkage.

makxenov: I thought about it, but now `proof-producer.nix` doesn't have access to all packages, and in this case it must have it to choose between `pkgs` and `pkgs.static`. So I will have to make something like `pkgs.callPackage ./proof-producer.nix { pkgs = pkgs; ... }` which seems weird to me.

proof-producer/CMakeLists.txt
akokoshn: We can don't use shared libraries at all

proof-producer/bin/proof-producer/CMakeLists.txt
makxenov: Unfortunately cmake doesn't have clear documentation for `target_link_libraries` either :( Seems it also accepts both link flags and libraries in LINK_OPTIONS section. So no differences to me.

flake.nix
makxenov: Konstantin mentioned ASan build where it's helpful to have shared libraries (can't say why, I am not an expert in sanitizers).

akokoshn: yes, according our dissolution, let's keep both build options: static is main (product), dynamic for additional checks

proof-producer/CMakeLists.txt
makxenov: Added error code to comment

AndreyMlashkin (approved):

AndreyMlashkin:
I would create a follow-up with steps like:
- make use of mold or gold for fast static linkage
- use only static binaries and drop support for dynamic (I am not talking about stdlib - it is still important to have it dynamically for sanitizers and other profiilng tools)

proof-producer/CMakeLists.txt
```diff
@@ -20,14 +20,24 @@ include(CMConfig)
 include(CMDeploy)
 include(CMSetupVersion)
 
+option(PROOF_PRODUCER_STATIC_BINARIES "Link proof-producer binaries statically" OFF)
+
+if(PROOF_PRODUCER_STATIC_BINARIES)
+    set(BUILD_SHARED_LIBS OFF)
+    set(Boost_USE_STATIC_LIBS ON)
+else()
+    set(BUILD_SHARED_LIBS OFF)  # TODO: fix invalid-pch issue with precompiled headers to build shared libraries
```
akokoshn, line 29: build dynamic assigner library works ok?

makxenov: Yes, the problem is with linking of binaries.

akokoshn (approved):

PR #226, opened 2024-12-23
assigner: add trace match check

oclaw:
Added to trace protobuf randomly generated value from trace_idx in order to check if all trace files passed to assigner are compatible to each other. Added option to skip this check if needed (not used now) + unit test with inconsistent data.

Old traces are not required to be re-generated, changes are not breaking

Also cleaned up assigner a bit


proof-producer/libs/assigner/include/nil/proof-generator/assigner/bytecode.hpp
```diff
@@ -8,15 +8,18 @@
 #include <nil/crypto3/zk/snark/arithmetization/plonk/assignment.hpp>
 #include <nil/blueprint/zkevm_bbf/bytecode.hpp>
 #include <nil/proof-generator/assigner/trace_parser.hpp>
+#include <nil/proof-generator/assigner/options.hpp>
 #include <nil/proof-generator/preset/limits.hpp>
+#include "nil/proof-generator/assigner/options.hpp"
```
akokoshn, line 13: duplicate?

proof-producer/libs/assigner/include/nil/proof-generator/assigner/copy.hpp
```diff
@@ -9,8 +9,10 @@
 
 #include <nil/crypto3/zk/snark/arithmetization/plonk/assignment.hpp>
 #include <nil/blueprint/zkevm_bbf/copy.hpp>
+#include <nil/proof-generator/assigner/options.hpp>
 #include <nil/proof-generator/assigner/trace_parser.hpp>
 #include <nil/proof-generator/preset/limits.hpp>
+#include "nil/proof-generator/assigner/options.hpp"
```
akokoshn, line 15: duplicate?

proof-producer/libs/assigner/include/nil/proof-generator/assigner/trace_parser.hpp
```diff
@@ -137,11 +141,46 @@ namespace nil {
...
+        class trace_index_mismatch: public std::exception {        
+        public:
+            [[nodiscard]] const char* what() const noexcept override {
+                return "Invalid trace index";
+            }
+        };
+
+        inline void check_trace_index(TraceIndex base, TraceIndex index) {
+            if (base != TraceIndex{} && index != base) {
+                throw trace_index_mismatch();
```
akokoshn, line 158: why do you want to  introduce exceptions here?
may be just return `nullopt` from deserialization function?

oclaw: initially thought to handle the error in one place but at the middle of the way decided optionally ignore index mismatch. Rewrote with `std::nullopt` and warning log

proof-producer/libs/assigner/include/nil/proof-generator/assigner/bytecode.hpp
oclaw: some vscode autocompletions, will revise them
cleaned up

proof-producer/libs/assigner/include/nil/proof-generator/assigner/copy.hpp
oclaw: fixed

proof-producer/tests/bin/proof-producer/test_zkevm_bbf_circuits.cpp
```diff
@@ -95,3 +95,19 @@ INSTANTIATE_TEST_SUITE_P(MultiTxZkevm, ProverTests, ::testing::Values(Input{Mult
 // ./build/bin/nil wallet new --amount 100000000000000000
 // ./build/bin/nil wallet deploy ./SimpleStorage.bin --abi ./SimpleStorage.abi
 // for i in {1..100}; do ./build/bin/nil wallet send-message 0x0001443b27c605cc3975e05e8eb68eef1913777c increment --abi ./SimpleStorage.abi --no-wait; done
+
+
+// RW trace is picked from another trace set and has different trace_idx
```
makxenov, line 100: ZKEVM trace actually

proof-producer/libs/assigner/include/nil/proof-generator/assigner/options.hpp
```diff
@@ -0,0 +1,22 @@
+#ifndef PROOF_GENERATOR_LIBS_ASSIGNER_OPTIONS_HPP_
+#define PROOF_GENERATOR_LIBS_ASSIGNER_OPTIONS_HPP_
+
+namespace nil {
+    namespace proof_generator {
+
+        struct AssignerOptions {
+            bool ignore_index_mismatch{};
```
makxenov, line 8: Why may we need to ignore the index?

akokoshn (approved):

proof-producer/tests/bin/proof-producer/test_zkevm_bbf_circuits.cpp
```diff
@@ -95,3 +95,19 @@ INSTANTIATE_TEST_SUITE_P(MultiTxZkevm, ProverTests, ::testing::Values(Input{Mult
...
+TEST(ProverTest, TraceIndexMismatch) {
+    const std::string trace_base_path = std::string(TEST_DATA_DIR) + "increment_broken_data.pb";
```
makxenov, line 102: Let's put broken data into separate folder inside `TEST_DATA_DIR`

proof-producer/libs/assigner/include/nil/proof-generator/assigner/options.hpp
oclaw: I am not sure actually, may be no need for this

But if there is some development or production use case that will require to pass into tracer binary compatible files from the same block but from different collection occurrences (let's say different kinds of traces were generated by separate collector nodes, dunno)  - IMO it is better to have such an option (disabled by default)

proof-producer/tests/bin/proof-producer/test_zkevm_bbf_circuits.cpp
oclaw: Nope, this test calls ZKEVM circuit. But it needs an RW operations files too for lookups generation and exactly this .rw file was spoofed in this trace set

proof-producer/tests/bin/proof-producer/test_zkevm_bbf_circuits.cpp
oclaw: done

proof-producer/libs/assigner/include/nil/proof-generator/assigner/options.hpp
makxenov: OK, but this logic with skipping zero index seems a bit confusing to me. How about passing skip flag explicitly and checking it inside `check_trace_index`?

P.S. I would prefer to avoid these code complications at all. It's not a big deal to regenerate trace files with same indices for development needs. Also this flag requires additional cmd option to be able to enable it. And I hardly can imagine that we will support this option in the caller of proof-producer (synccommittee) to use skipping in production.

makxenov (approved):

proof-producer/libs/assigner/include/nil/proof-generator/assigner/options.hpp
oclaw: I designed it like this to be independent from call order and be able to pass zero index in case if we do not have its value yet

But you are right, it is a bit unclear. Now I am passing AssignerOptions to deserializers and base_index is just optional for explicitness

x-mass (approved):

PR #230, opened 2025-01-02
Check proto file hash in trace files

makxenov:


akokoshn (approved):

akokoshn (approved):

proof-producer/libs/assigner/CMakeLists.txt
```diff
@@ -23,13 +23,23 @@ foreach(PROTO_FILE ${PROTO_FILES})
     list(APPEND GENERATED_HEADERS ${PROTO_HDR})
 endforeach()
 
+# Write header file with version of trace proto spec
+set(TRACE_PROTO_FILE ${CMAKE_CURRENT_SOURCE_DIR}/proto/trace.proto)
+set(PROTO_HASH_HEADER "${CMAKE_CURRENT_BINARY_DIR}/proto_hash.h")
+add_custom_command(
```
oclaw, line 29: I would use [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html) with predefined header file `proto_hash.h.in` file instead of assembling it each time from bash output. It has more guarantees and easier to read

oclaw: Agreed on this approach to enforce hash recalculation on each build attempt

oclaw (approved):

Issue #125, opened 2024-10-30
Placeholder with KZG commitment scheme fails on circuits with lookups


vo-nil:
Assertion is triggered here:
https://github.com/NilFoundation/placeholder/blob/master/crypto3/libs/zk/include/nil/crypto3/zk/snark/systems/plonk/placeholder/lookup_argument.hpp#L618

PR #235, opened 2025-01-03
Add test with lookup_argument to KZG commitment scheme

vo-nil:


vo-nil:
This test fails with assert from #125

PR #251, opened 2025-01-08
write ctest console output to the file and display it to the log; temporaly disable invalid tests

AndreyMlashkin:
Display ctest console output  life, as it may be needed, when OOM happens


AndreyMlashkin:
This is a follow up to re-enable the tests:
https://github.com/NilFoundation/placeholder/issues/252

PR #258, opened 2025-01-10
proof-producer: introduce command chain

oclaw:
Implemented resource subscription mechanism based on boost::signals + command_chain for achieving smart resource management. Now each step of the executing command keeps its own copy of shared pointer to the data it uses

Rewrote all prover command to the new syntax. Now each command is located into its own file and represented as queue of executing steps (each step releases references to the resources after popping from the queue)

New approach allows:
- keep different parts of proof-producer independently
- lazy-initialize large objects (LpcScheme, AssignmentTable)
- release these resources automatically based on the steps added to the command

Related changes in crypto3:
- common_data is now placed into shared_ptr as far as it is used as an independent data part in most cases
- Added reset of KZG scheme in unit tests

Also added ResultCode types for identifying the exact reason of failure externally


proof-producer/bin/proof-producer/include/nil/proof-generator/type_system.hpp
oclaw, line 1: Aliases from Prover that are going to be shared between commands

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
```diff
@@ -0,0 +1,264 @@
...
+
+            struct BinaryWriter: public command_step {
+
+                BinaryWriter(
+                    resources::resource_provider<AssignmentTable>& table_provider,
+                    resources::resource_provider<TableDescription>& description_provider,
+                    const boost::filesystem::path& output_filename): 
+                output_filename_ (output_filename) 
+                {
+                    resources::subscribe_value<AssignmentTable>(table_provider, assignment_table_);
```
oclaw, line 131: Subscription to resource example (something close to Qt signal-slot system)

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
```diff
@@ -0,0 +1,264 @@
...
+            };
+
+            PresetCommand(const Args& args) {
+
+                using PresetStep = typename PresetStep<CurveType, HashType>::Executor;
+                using CircuitWriteStep = typename CircuitWriteStep<CurveType, HashType>::Executor;
+                using AssignmentTableBinaryWriter = typename AssignmentTableWriteStep<CurveType, HashType>::BinaryWriter;
+                using AssignmentTableDebugPrinter = typename AssignmentTableWriteStep<CurveType, HashType>::DebugPrinter;
+
+                auto& circuit_maker = add_step<PresetStep>(args.circuit_name); // init circuit for the given name
```
oclaw, line 243: Adding step to the chain example

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
```diff
@@ -0,0 +1,264 @@
...
+                }
+
+            private:
+                const boost::filesystem::path output_filename_;
+                std::shared_ptr<AssignmentTable> assignment_table_;
+                std::shared_ptr<TableDescription> table_description_;
+            };
+
+
+            struct DebugPrinter: public command_step {
```
oclaw, line 165: Command step example: isolated functionality part with required resources

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
```diff
@@ -0,0 +1,264 @@
...
+            using TableDescription = typename Types::TableDescription;
+
+            struct Executor: 
+                public command_step, 
+                public resources::resources_provider<ConstraintSystem, AssignmentTable, TableDescription>
+            {
+
+                Executor(const std::string& circuit_name): circuit_name_(circuit_name) {}
+
+                bool execute() override
```
oclaw, line 35: `Prover::setup_prover()`

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
```diff
@@ -0,0 +1,264 @@
...
+            struct Args { // TODO fill from boost program options and print as help
+                std::string circuit_name;
+                boost::filesystem::path circuit_file_path;
+                boost::filesystem::path assignment_table_file_path;
+                nil::proof_generator::OutputArtifacts output_artifacts;                
+            };
+
+            PresetCommand(const Args& args) {
+
+                using PresetStep = typename PresetStep<CurveType, HashType>::Executor;
```
makxenov, line 238: Why do we need nested `Executor` here?

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,61 @@
...
+    command_step& operator=(command_step&&) = default;
+};
+
+class command_chain: public command_step {
+
+public:
+    bool execute() override {
+        while (!steps_.empty()) {
+            if (!steps_.front()->execute()) {
+                return false;
```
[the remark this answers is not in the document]
oclaw: I thought to check execution result of the whole command in main() and print an error if execution result is false
But you are right, it may lead to silent "command failed without an explanation"

I can replace bool with `using CommandResult = std::optional<std::string>` to force developer to return readable error from the command, what do you think?

proof-producer/bin/proof-producer/include/nil/proof-generator/resources.hpp
```diff
@@ -0,0 +1,68 @@
...
+        resource_provider& operator=(resource_provider&&) = default;
+
+        Signal sig_impl_; // TODO make me private
+    };
+
+    template <typename... Resources> 
+    class resources_provider: public resource_provider<Resources>... {};
+
+    template <typename T, typename... Resources> 
+    concept ProvidesAll = (std::is_base_of_v<resource_provider<Resources>, T> && ...);
```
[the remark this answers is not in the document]
oclaw: It still might be used, will remove if there will be no need for it

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
oclaw: Because of or cumbersome template system :)
I want to use human-readable shortened names inside code of the command. And to do so I have to declare my class with command implementation in some scope already parametrized with `<CurveType, HashType>` and something else in future maybe. So this external classes are responsible for template specialization and Executor(s) (in other parts there be several of them) contain the exact command step logic

proof-producer/bin/proof-producer/include/nil/proof-generator/resources.hpp
```diff
@@ -0,0 +1,68 @@
...
+        resource_provider(const resource_provider&) = delete;
+        resource_provider& operator=(const resource_provider&) = delete;
+        resource_provider(resource_provider&&) = default;
+        resource_provider& operator=(resource_provider&&) = default;
+
+        Signal sig_impl_; // TODO make me private
+    };
+
+    template <typename... Resources> 
+    class resources_provider: public resource_provider<Resources>... {};
```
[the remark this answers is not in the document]
oclaw: Actually there is just an alias to simplify inheritance syntax, not sure if there is need to distinct them

proof-producer/libs/preset/include/nil/proof-generator/preset/types.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <cstdint>
+
+#include <nil/blueprint/blueprint/plonk/circuit.hpp>
+#include <nil/crypto3/zk/snark/arithmetization/plonk/assignment.hpp>
+
+namespace nil {
+    namespace proof_generator {
+
+        template <typename BlueprintFieldType>
+        struct PresetTypes {
```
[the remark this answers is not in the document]
oclaw: That's the code structure question. Code parts for actual circuit initialization and assignment table generation are located in the independent libraries so right now I cannot reuse header from proof-producer executable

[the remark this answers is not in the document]
oclaw: The only idea I can suggest is to isolate whole type system into another interface library and reuse it among all proof-producer parts (both binary and libraries), it looks reasonable
If you agree on that - I will do this after I finish logic parts refactoring

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <utility>
+
+namespace nil {
+
+    namespace proof_generator {
+
+
+// basic interface for a command step
+// non-copyable as it is designed to be used only as part of a command chain
+// each step should keep pointers to the resources it needs and accept them (or them providers) in the constructor
```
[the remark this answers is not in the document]
oclaw: yep, will fix

proof-producer/libs/preset/include/nil/proof-generator/preset/types.hpp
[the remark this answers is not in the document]
makxenov: I see. Yes, it's a good idea

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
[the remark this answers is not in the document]
makxenov: I would vote for more convenient `ProverError` struct with explicit static methods for construction (for example, `return error::NoError();`/`return errors::MakeError("...");`)

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
oclaw: It is a bit more complicated unfortunately, I have 2 reasons

1. For code parts that are working on the same type subset but implementing different steps there is convenient to have shared type aliases. For example I want to keep different AssignmentTable binary/text readers/writers into the same scope as far as they use common type aliases and

2. For steps that are derived classes of `resource_provider<T>` attempt to declare them without nesting will lead to a construction like
```
template<typename CurveType, typename HashType>
class Assigner: public resources::resource_provider<TypeSystem::AssignmentTable<CurveType, HashType>> {
    using Types = TypeSystem<CurveType, HashType>;
    using AssignmentTable = Types::AssignmentTable; // we need this alias on the first line but cannot obtain it
    ....
};
```

I agree that nesting is not so obvious too, but this is the least ugly option came into my head :)

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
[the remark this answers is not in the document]
oclaw: Yep, that's an option too :)

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
```diff
@@ -0,0 +1,264 @@
...
+#include <nil/proof-generator/output_artifacts/circuit_writer.hpp>
+#include <nil/proof-generator/output_artifacts/output_artifacts.hpp>
+#include "nil/proof-generator/output_artifacts/output_artifacts.hpp"
+
+namespace nil {
+    namespace proof_generator {
+
+        template<typename CurveType, typename HashType>
+        struct PresetStep {
+            using Types            = TypeSystem<CurveType, HashType>;
```
akokoshn, line 22: let' avoid duplication `using` in each method

proof-producer/bin/proof-producer/include/nil/proof-generator/resources.hpp
```diff
@@ -0,0 +1,68 @@
...
+        using Signal = boost::signals2::signal<void(ResourcePtr)>;
+        using SlotType = Signal::slot_type;
+
+        resource_provider() = default;
+        resource_provider(const resource_provider&) = delete;
+        resource_provider& operator=(const resource_provider&) = delete;
+        resource_provider(resource_provider&&) = default;
+        resource_provider& operator=(resource_provider&&) = default;
+
+        Signal sig_impl_; // TODO make me private
```
akokoshn, line 25: why we actually need notification mechanism here?

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
```diff
@@ -0,0 +1,264 @@
...
+            using BlueprintField   = typename Types::BlueprintField;
+            using ConstraintSystem = typename Types::ConstraintSystem;
+            using AssignmentTable  = typename Types::AssignmentTable;
+            using TableDescription = typename Types::TableDescription;
+
+            struct Executor:
```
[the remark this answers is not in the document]
oclaw: Discussed [here](https://github.com/NilFoundation/placeholder/pull/258#discussion_r1914812846)

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
oclaw: Actually this is the one of the points I wanted to simplify during the refactoring. I have two thoughts:
- We have to keep the whole set of type aliases from crypto3 in one point - otherwise it will be hard to find common patterns and primitives between different commands

- The whole type system is very big and hard to read - so I decided to use some import-like mechanism to grab only needed for the command types in nice human-readable way

I can switch to direct usage of templated types, but I do not see why it is better tbh

proof-producer/bin/proof-producer/include/nil/proof-generator/resources.hpp
oclaw: Because of 2 points:

- Each command step might provide some data to the next ones. These data objects (AssignmentTable, LpcScheme) might be huge and we want to get rid of them as soon as they are not required anymore
- The exact moment of initialization of these objects is unknown by the command start, resource may be initialized on 3rd or 5th step or not initialized at all (e.g. depending on the CLI options)

This issue could be solved with:
 1. Spreading references to some `shared_ptr<Object>` declared in the command scheduler between all of the steps. It will require hacks like resetting pointer to some value during the initialization and may lead to resource leaks because of keeping unneeded references to the resource because of inaccurate copies
 
 2. Spreading these shared pointers only at the exact initialization point of execution via publisher-subscriber pattern, this is where this notification mechanism comes into scene. IMO this approach has more explicit semantics of usage: 
 - you actually see through code which step provides the resource and which ones consume it
 - it may be enriched with some additional code safety helpers through modifying `subscribe/notify` methods (for example for one-time initialization guarantee)
 - code of the constructors is actually not so different between using plain shared_pointers and resource providers, both have one-line initialization in most cases

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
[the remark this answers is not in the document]
oclaw: Or you meant to keep consistency between `struct vs class` definition? I got used to a pattern when abstract types are declared with `class` keyword for some reason :)
I can re-declare all of the nested types as classes, this is not a big deal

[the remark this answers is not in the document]
akokoshn: Yes, i mostly about consistency. Generally ok with current design - split abstract interfaces and implementation like class/struct.

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
akokoshn: make sense, but redeclare types in each method/function - so many duplicates.
May be declare once for logical module?

oclaw: That's what I am trying to achieve, yes. Pushing updated code structure soon, let's review it again later

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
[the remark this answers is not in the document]
oclaw: After discussion with @zadykian we it seems even more useful. Added types ResultCode and CommandResult for storing some determined err code and human readable description. Migrated commands to use this type and going to add a ticket for using proper ret code in each return from the prover

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
oclaw: Finished splitting functionality to files, you can check how do you like it now :)

crypto3/libs/transpiler/test/evm.cpp
```diff
@@ -1389,8 +1389,9 @@ BOOST_FIXTURE_TEST_CASE(transpiler_test, test_tools::random_test_initializer<fie
             output + "/" + public_input_filename
         );
 
+        kzg_scheme = kzg_scheme_type(kzg_params);
```
oclaw, line 1392: Without this reset KZG tests failing (LPC tests already do this)

proof-producer/bin/proof-producer/include/nil/proof-generator/preset_command.hpp
akokoshn: Yes, `TypeSystem<...>` looks much better)

proof-producer/bin/proof-producer/include/nil/proof-generator/resources.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+        resource_provider() = default;
+        resource_provider(const resource_provider&) = delete;
+        resource_provider& operator=(const resource_provider&) = delete;
+        resource_provider(resource_provider&&) = default;
+        resource_provider& operator=(resource_provider&&) = default;
+        
+        virtual ~resource_provider() = default;
+
+        Signal sig_impl_; // TODO make me private
```
makxenov, line 26: Any difficulties with this TODO?

proof-producer/bin/proof-producer/src/main.cpp
```diff
@@ -17,184 +17,263 @@
...
+#include <nil/proof-generator/commands/verify_command.hpp>
+#include <nil/proof-generator/commands/all_command.hpp>
+#include <nil/proof-generator/commands/agg_challenge_command.hpp>
+#include <nil/proof-generator/commands/merge_proofs_command.hpp>
+#include <nil/proof-generator/commands/compute_combined_q_command.hpp>
+#include <nil/proof-generator/commands/aggregated_fri_proof_command.hpp>
+#include <nil/proof-generator/commands/gen_consistency_check_command.hpp>
+
+
+
```
makxenov, line 40: Too many newlines

makxenov:
General question. We have 2 similar terms: proof producer (as top-level directory, Nix derivation name, name of binaries) and proof generator (namespace for source files inside `proof-producer/`, name of libraries). Do these terms have any differences in the meaning? If no, let's unify it to avoid possible confusions. 
@akokoshn @oclaw @x-mass

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,145 @@
+#ifndef PROOF_GENERATOR_ASSIGNER_COMMAND_STEP_HPP
```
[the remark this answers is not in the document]
oclaw: I totally agree but I do not observe consistency in our repo unfortunately
If there is no other opinions - I will gladly replace all of these guards with # pragma once

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,145 @@
...
+            constexpr static CommandResult Ok() noexcept {
+                return CommandResult();
+            }
+            
+            template <typename... Args>
+            static CommandResult Error(ResultCode rc, std::format_string<Args...> _fmt, Args&&... args) {
+                return CommandResult(rc, std::format(_fmt, std::forward<Args>(args)...));
+            }
+
+            // TODO Remove this function and use the Error function instead
```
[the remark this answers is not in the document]
oclaw: Yes, have already created a ticket for that

proof-producer/bin/proof-producer/CMakeLists.txt
```diff
@@ -66,8 +66,11 @@ function(setup_proof_generator_target)
     target_link_libraries(${ARG_TARGET_NAME} PRIVATE
         ${INTERFACE_LIBS}
         Boost::program_options
+        proof_generatorTypes
```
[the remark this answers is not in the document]
oclaw: I am totally agree. @akokoshn  as an introducer of the current format, what do you think?

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,145 @@
...
+            constexpr ResultCode result_code() const noexcept {
+                return result_;
+            }
+
+            std::string error_message() const {
+                if (succeeded()) {
+                    return "";
+                }
+
+                return std::format("result_code={}({}): {}",
```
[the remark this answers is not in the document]
oclaw: Not sure. 
Most probably the process of the debugging of some issue with the proof-producer will lead to search for the exact error code in sources, so IMO presence of the numeric error code value in each log will make it easier. We may introduce another method containing only the description but I barely can imagine its usecase tbh.

proof-producer/libs/types/include/nil/proof-generator/types/type_system.hpp
```diff
@@ -0,0 +1,103 @@
...
+        };
+
+        // common type definitions needed for proof generation & verification steps
+        template<typename CurveType, typename HashType>
+        struct TypeSystem {
+            using BlueprintField = typename CurveType::base_field_type;
+            using LpcParams = nil::crypto3::zk::commitments::list_polynomial_commitment_params<HashType, HashType, 2>;
+            using Lpc = nil::crypto3::zk::commitments::list_polynomial_commitment<BlueprintField, LpcParams>;
+            using LpcScheme = typename nil::crypto3::zk::commitments::lpc_commitment_scheme<Lpc>;
+            using polynomial_type = typename LpcScheme::polynomial_type;
```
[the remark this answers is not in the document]
oclaw: Seems like some late addition to the type system. Will clean it up a bit later, but current polynomial_io already uses PolynomialType with different semantics, that is the only reason I have not replaced it earlier. Maybe @martun has some thoughts of it :)

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/gen_proof_command.hpp
```diff
@@ -0,0 +1,473 @@
...
+                }
+
+            private:
+                boost::filesystem::path proof_file_;    
+            };
+            
+            struct ProofGenerator: 
+                public ProofGeneratorBase,
+                public command_step,
+                public resources::resource_provider<Proof>
```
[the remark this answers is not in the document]
oclaw: It's in verifier

proof-producer/bin/proof-producer/CMakeLists.txt
```diff
@@ -100,3 +103,4 @@ install(TARGETS ${MULTI_THREADED_TARGET} RUNTIME DESTINATION bin)
 
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+
```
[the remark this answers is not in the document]
oclaw: done

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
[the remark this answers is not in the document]
oclaw: done

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,145 @@
...
+#include <boost/log/trivial.hpp>
+
+
+namespace nil {
+
+    namespace proof_generator {
+
+        // TODO use these in commands and command steps
+        enum class ResultCode: uint8_t {
+            None = 0,
```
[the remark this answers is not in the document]
oclaw: agree, renamed

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/agg_challenge_command.hpp
```diff
@@ -0,0 +1,87 @@
...
+namespace nil {
+    namespace proof_generator {
+
+        template <typename CurveType, typename HashType>
+        struct AggregatedChallengeCommand: public command_step {
+
+            using Types                  = TypeSystem<CurveType, HashType>;
+            using BlueprintField         = typename Types::BlueprintField;
+            using PlaceholderParams      = typename Types::PlaceholderParams;
+            using TTypeBase              = typename Types::TTypeBase;
```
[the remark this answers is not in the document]
oclaw: cleaned up

proof-producer/bin/proof-producer/src/arg_parser.cpp
```diff
@@ -126,6 +126,7 @@ namespace nil {
             po::options_description cmdline_options("nil; Proof Producer");
             cmdline_options.add(generic).add(config);
 
+
```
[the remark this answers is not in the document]
oclaw: cleaned up

proof-producer/bin/proof-producer/src/main.cpp
```diff
@@ -17,184 +17,263 @@
...
                     break;
+                }
                 case nil::proof_generator::detail::ProverStage::FAST_GENERATE_PARTIAL_PROOF:
-                    // Preset, fill assignment table, preprocess
-                    prover_result =
-                        prover.setup_prover() &&
-                        prover.fill_assignment_table(prover_options.trace_base_path) &&
-                        prover.save_assignment_description(prover_options.assignment_description_file_path) &&
-                        prover.preprocess_public_data() &&
-                        prover.save_preprocessed_common_data_to_file(prover_options.preprocessed_common_data_path) &&
```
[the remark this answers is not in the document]
oclaw: fixed, thanks!

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/preprocess_command.hpp
```diff
@@ -0,0 +1,231 @@
...
+#include <nil/proof-generator/output_artifacts/output_artifacts.hpp>
+
+#include <nil/crypto3/zk/snark/systems/plonk/placeholder/preprocessor.hpp>
+
+
+namespace nil {
+    namespace proof_generator {
+
+        template <typename CurveType, typename HashType>
+        struct CommitmentSchemeFactory {
```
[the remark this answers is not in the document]
oclaw: did for this class and for the prover-related stuff

proof-producer/bin/proof-producer/src/main.cpp
oclaw: cleaned up

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/gen_proof_command.hpp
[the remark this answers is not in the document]
makxenov: Thanks, I see

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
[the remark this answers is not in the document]
makxenov: Yes, I agree that a sole description would be useless while we have this method. My point is that I expect from method named as the field to be a simple getter, but the current implementation is something else. That's why I suggested to rename it.

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/detail/commitment_scheme_factory.hpp
```diff
@@ -0,0 +1,30 @@
...
+                std::size_t table_rows_log = std::ceil(std::log2(rows_amount));
+
+                return std::make_shared<LpcScheme>(LpcScheme(FriParams(1, table_rows_log, 
+                    config_.lambda, config_.expand_factor, config_.grind!=0, config_.grind)));
+            }
+
+            const PlaceholderConfig config_;
+        };
+    } // namespace proof_generator
+} // namespace nil
```
makxenov, line 30: Seems the last one newline isssue :)

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/detail/proof_gen.hpp
```diff
@@ -0,0 +1,288 @@
...
+            {
+                PartialProofGenerator(
+                    resources::resource_provider<ConstraintSystem>& constraint_provider,
+                    resources::resource_provider<AssignmentTable>& table_provider,
+                    resources::resource_provider<TableDescription>& desc_provider,
+                    resources::resource_provider<PublicPreprocessedData>& public_data_provider,
+                    resources::resource_provider<LpcScheme>& lpc_scheme_provider,
+                    resources::resource_provider<PrivatePreprocessedData>& private_data_provider,
+                    const boost::filesystem::path& proof_file,
+                    const boost::filesystem::path& challenge_file_,
```
makxenov, line 190: Minor: final underscore in argument variable name

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/gen_proof_command.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <nil/proof-generator/commands/detail/io/lpc_scheme_io.hpp>
+#include <nil/proof-generator/commands/preprocess_command.hpp>
+#include <nil/proof-generator/commands/detail/proof_gen.hpp>
+#include <nil/proof-generator/commands/fill_assignment_command.hpp>
+#include <nil/proof-generator/output_artifacts/output_artifacts.hpp>
+
+namespace nil {
+    namespace proof_generator {
+
+        // TODO move to files
```
makxenov, line 22: This todo is resolved, isn't it?

proof-producer/bin/proof-producer/src/arg_parser.cpp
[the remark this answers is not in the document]
makxenov: Some artifacts still remain :( 
P.S. if you are using VSCode by a chance, I strongly recommend to apply these settings: `"files.insertFinalNewline": true, "files.trimFinalNewlines": true, "files.trimTrailingWhitespace": true`, they work great

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/detail/commitment_scheme_factory.hpp
oclaw: fixed

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/detail/proof_gen.hpp
oclaw: fixed

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/gen_proof_command.hpp
oclaw: it is, thanks, cleaned up

oclaw:
Essentially I used them but some it looks like some VSCode update just threw them out ¯\_(ツ)_/¯
Restores them

proof-producer/bin/proof-producer/src/arg_parser.cpp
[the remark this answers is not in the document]
oclaw: Essentially I used them but it looks like some VSCode update just threw them out ¯_(ツ)_/¯
Restored them

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+        public:
+            CommandResult execute() override final {
+                int stage{1};
+                int total_stages = steps_.size();
+                while (!steps_.empty()) {
+                    auto const res = steps_.front()->execute();
+                    if (!res.succeeded()) 
+                    {
+                        BOOST_LOG_TRIVIAL(error) << "command failed on stage " << stage << " of " << total_stages << ": " << res.error_message();
```
akokoshn, line 117: nice to have print here name of the stage together with number

akokoshn:
> General question. We have 2 similar terms: proof producer (as top-level directory, Nix derivation name, name of binaries) and proof generator (namespace for source files inside `proof-producer/`, name of libraries). Do these terms have any differences in the meaning? If no, let's unify it to avoid possible confusions. @akokoshn @oclaw @x-mass

@martun , do you remember why was used `proof_generator` namespace but not `proof_producer`?

proof-producer/bin/proof-producer/src/main.cpp
```diff
@@ -17,184 +17,265 @@
...
-                        prover_result = prover.save_assignment_description(prover_options.assignment_description_file_path);
-                    }
+                {
+                    using Command = FillAssignmentCommand<CurveType, HashType>;
+                    using Args = typename Command::Args;
+                    Command cmd(Args{
+                        .circuit_name = prover_options.circuit_name,
+                        .trace_file_path = prover_options.trace_base_path,
+                        .circuit_file_path = prover_options.circuit_file_path,
+                        .assignment_table_file_path = prover_options.assignment_table_file_path,
```
makxenov, line 97: Some of the commands specify their output arguments with `out_` prefix, but other ones do not.

proof-producer/tests/bin/proof-producer/test_zkevm_bbf_circuits.cpp
```diff
@@ -23,39 +24,52 @@ class ProverTests: public ::testing::TestWithParam<Input> {
...
-        using BlueprintFieldType = nil::proof_generator::Prover<CurveType, HashType>::BlueprintField;
 
-        static constexpr std::size_t lambda = 9;
-        static constexpr std::size_t grind = 0;
-        static constexpr std::size_t expand_factor = 2;
-        static constexpr std::size_t max_quotient_chunks = 0;
+        using ConstraintSystem = nil::proof_generator::TypeSystem<CurveType, HashType>::ConstraintSystem;
+        using BlueprintFieldType = nil::proof_generator::TypeSystem<CurveType, HashType>::BlueprintField;
+        using AssignmentTable = nil::proof_generator::TypeSystem<CurveType, HashType>::AssignmentTable;
+        using FillAssignmentCommand = nil::proof_generator::FillAssignmentCommand<CurveType, HashType>;
```
akokoshn, line 31: seems not used

akokoshn (commented):
generally look good.
Let's wait for @martun  review

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,141 @@
...
+        enum class ResultCode: uint8_t {
+            Success = 0,
+            IOError = 10,        // cannot access some of input files
+            InvalidInput = 20,   // input files are inconsistent or malformed
+            ProverError = 30,    // some logical error from proof system
+            UnknownError = 0xFF,
+        };
+
+        inline constexpr const char* result_code_to_string(ResultCode rc) {
+            switch (rc) {
```
martun, line 27: The rest of the project is throwing exceptions. You will still need to have a "catch block" at the start of each command to report those exceptions if there will be any. Maybe here too you could just throw standard std::exceptions.

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/detail/commitment_scheme_factory.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <nil/proof-generator/types/type_system.hpp>
+
+
+namespace nil {
+    namespace proof_generator {
+
+        template <typename CurveType, typename HashType>
+        struct CommitmentSchemeFactory {
```
martun, line 10: Do we actually create more than 1 commitment scheme per run? Maybe just 2? Do we want a factory class for creating just 2 instances of a class?

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/agg_challenge_command.hpp
```diff
@@ -0,0 +1,82 @@
...
+            using Types                  = TypeSystem<CurveType, HashType>;
+            using BlueprintField         = typename Types::BlueprintField;
+            using PlaceholderParams      = typename Types::PlaceholderParams;
+
+            struct Args {
+                std::vector<boost::filesystem::path> aggregate_input_files;
+                boost::filesystem::path aggregated_challenge_file;
+            };
+
+            AggregatedChallengeCommand(const Args& args): args_(args) {}
```
martun, line 32: Do we use this "Args" structure somewhere else? Can't we just accept the arguments here, and delete the "Args" structure from everywhere? 

If we use the Args, I would suggest to mark somehow which arguments are inputs, which are the outputs, and when printing the error messages from the "execute" function of command_step.hpp, print the name of the command and the parameters it was called with as a part of the error message. This way will use the Args in some way.

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/agg_challenge_command.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+        private:
+            static CommandResult generate_aggregated_challenge_to_file(
+                    const std::vector<boost::filesystem::path> &aggregate_input_files,
+                    const boost::filesystem::path &aggregated_challenge_file
+                ) 
+            {
+                using ChallengeIO = ChallengeIO<CurveType, HashType>;
+
+                if (aggregate_input_files.empty()) {
```
martun, line 52: Please consider moving this and similar checks to the constructor of this class. It's just a suggestion, it's ok here as well.

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/aggregated_fri_proof_command.hpp
```diff
@@ -0,0 +1,203 @@
...
+#include <nil/crypto3/marshalling/math/types/polynomial.hpp>
+#include "nil/proof-generator/commands/detail/io/polynomial_io.hpp"
+
+
+namespace nil {
+    namespace proof_generator {
+
+            template <typename CurveType, typename HashType>
+            struct AggregatedFriProofGenerator: public command_step
+             {
```
martun, line 29: You can move the scope { to the previous line.

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/aggregated_fri_proof_command.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+                boost::filesystem::path aggregated_challenge_file_;
+                std::vector<boost::filesystem::path> input_combined_Q_polynomial_files_;
+                boost::filesystem::path aggregated_fri_proof_output_file_;
+                boost::filesystem::path proof_of_work_output_file_;
+                boost::filesystem::path consistency_checks_challenges_output_file_;
+            };
+
+        template <typename CurveType, typename HashType>
+        struct AggregatedFriProofCommand: public command_chain {
```
martun, line 171: Some commands derive from chain, some others from command_step itself. Maybe we derive them all from chain?

I think if we want to be extremely picky, we could create not a chain but a tree of commands, so each command will know its subcommands which it needs to perform to finish the operation, something like the "Composite Pattern".

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/all_command.hpp
```diff
@@ -0,0 +1,110 @@
...
+                    table_reader, 
+                    table_reader, 
+                    circuit_reader
+                );   
+                auto& private_preprocessor = add_step<PrivatePreprocessor>(circuit_reader, table_reader, table_reader); // preprocess private data
+                auto& prover = add_step<Prover>(                                                                             // generate proof
+                    circuit_reader,
+                    table_reader,           // for table    
+                    table_reader,           // for table description
+                    public_preprocessor,    // for public data
```
martun, line 74: It looks a bit strange that we pass the same object twice, because it contains inside it 2 object that class "Prover" will need to access. Why not extract those objects (the public data and LPC scheme) and pass them to this class? The benefit of creating class "resources::resource_provider" is to load the resource from a file only when required? Is it worth it writing so much code for that purpose?

martun (approved):

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/agg_challenge_command.hpp
oclaw: Separate args structs will help us to split command line options in future. I thought to do it during the initial refactoring too but amount of LoC made me to reconsider :)

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,141 @@
...
+                    stage++;
+                }
+                return CommandResult::Ok();
+            }
+
+        protected:
+            // returns reference to the pushed step (non-owning, ownership is guaranteed by the step queue)
+            template <typename Step, typename... Args>
+                requires (std::derived_from<Step, command_step> && std::constructible_from<Step, Args...>) 
+            Step& add_step(Args&&... args) {
```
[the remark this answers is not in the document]
oclaw: It is complicated only here and the pattern is much similar to `make_shared` . IMO such approach reduces possible resource leaks caused by unneeded copies of command_step instances outside the queue

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
```diff
@@ -0,0 +1,141 @@
...
+            template <typename Step, typename... Args>
+                requires (std::derived_from<Step, command_step> && std::constructible_from<Step, Args...>) 
+            Step& add_step(Args&&... args) { 
+                BOOST_LOG_TRIVIAL(trace) << "adding " << steps_.size() + 1 << " step: " << __PRETTY_FUNCTION__;
+                std::unique_ptr<command_step>& pushed = steps_.emplace(std::make_unique<Step>(std::forward<Args>(args)...));
+                return dynamic_cast<Step&>(*pushed); // safe here because we know that we just pushed exactly this type
+            }
+
+        private:
+            std::queue<std::unique_ptr<command_step>> steps_;
```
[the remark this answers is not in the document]
oclaw: Discussed in DM

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
oclaw: Last time I saw entire proof-producer is not throwing exceptions, maybe your take is more relevant to the crypto3 internals

I, actually, do not mind switching to an exception-based mechanism of error handling, but it is a topic for a separate discussion
cc @akokoshn

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/agg_challenge_command.hpp
oclaw: Yes, I will try to make consistent args checking, these are the part of migrated and unchanged code parts

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/all_command.hpp
oclaw: This is the part I am not proud of too :(

I think it could be improved with some another factory methods or some kind of dependency injection but it will bring much more complications to an already complicated logic

proof-producer/bin/proof-producer/CMakeLists.txt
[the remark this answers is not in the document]
oclaw: done

proof-producer/bin/proof-producer/src/main.cpp
oclaw: made it consistent

proof-producer/bin/proof-producer/include/nil/proof-generator/commands/detail/commitment_scheme_factory.hpp
oclaw: It is just a wrapper for initializing lpc scheme in a same way in different stages

proof-producer/bin/proof-producer/src/arg_parser.cpp
```diff
@@ -139,7 +139,7 @@ namespace nil {
             // clang-format on
             po::options_description cmdline_options("nil; Proof Producer");
             cmdline_options.add(generic).add(config);
-
+
```
makxenov, line 142: Nitpick: file still has trailing whitespaces

makxenov (approved):

proof-producer/bin/proof-producer/src/arg_parser.cpp
oclaw: I actually wanted to include 'clang-format' run to this PR but it would turn the whole diff to a mess
I suggest to create a precommit hook with implicit call of it in further development :)

akokoshn (approved):

martun:
> > General question. We have 2 similar terms: proof producer (as top-level directory, Nix derivation name, name of binaries) and proof generator (namespace for source files inside `proof-producer/`, name of libraries). Do these terms have any differences in the meaning? If no, let's unify it to avoid possible confusions. @akokoshn @oclaw @x-mass
> 
> @martun , do you remember why was used `proof_generator` namespace but not `proof_producer`?

I think that's an accident, "proof generator" and "proof producer" are the same, you can change it to proof producer.

proof-producer/bin/proof-producer/include/nil/proof-generator/command_step.hpp
oclaw: I am doing something like that with __PRETTY_FUNCTION__ macro usage in `add_step<typename Step>`
Its output is a bit redundant but it gives the context. Maybe later I will prettify it additionally

makxenov (approved):

PR #265, opened 2025-01-14
Blueprint disable unnecessary test and mac platform bug fixes.

Iluvmagick:
immitrin.h include fix on non-intel
current check is

__has_include(<immintrin.h>)
but in practice nix does pull that one even on non-intel systems
i stole the fix from <immintrin.h> itself, as it checks that the platform is good inside
additionally

#include <bits/endian.h>
does not exist on darwin; switched to boost's detection methods for this one

also fixed a mac compiler eccentricity, which required explicitly specifying a template argument in one case i found

additionally, removed testing stretched component in blueprint_algebra_fields_plonk_non_native_multiplication_test, as this functionality is currently unused and the test fails in CI (although it is fine on my machine for some reason).

### DIFF
--- a/crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/detail/intel_intrinsics.hpp
+++ b/crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/detail/intel_intrinsics.hpp
@@ -14,7 +14,9 @@
 #include "nil/crypto3/multiprecision/detail/config.hpp"
 #include "nil/crypto3/multiprecision/detail/int128.hpp"
 
-#if __has_include(<immintrin.h>)
+// nix will sometimes have immintrin.h even on non-x86 platforms
+// so we additionally check for x86_64 or i386 (using the same checks as in immintrin.h)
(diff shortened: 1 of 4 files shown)


martun (approved):

crypto3/libs/multiprecision/include/nil/crypto3/multiprecision/detail/intel_intrinsics.hpp
```diff
@@ -55,7 +55,27 @@ namespace nil::crypto3::multiprecision::detail {
...
 
 #endif
+}  // namespace nil::crypto3::multiprecision::detail
+
+#else
+#warning "Missing immintrin.h, addcarry and subborrow optimizations disabled"
+
+namespace nil::crypto3::multiprecision::detail {
+
+    NIL_CO3_MP_FORCEINLINE unsigned char addcarry_limb(unsigned char carry, limb_type a,
```
ioxid, line 65: We don't need this fallback. We can keep this for now, I will fix it in my PR.

ioxid (approved):

PR #269, opened 2025-01-15
Do not save common data on fast partial proof

makxenov:
Preprocessed common data is not used in further dFRI steps

### DIFF
--- a/proof-producer/bin/proof-producer/src/main.cpp
+++ b/proof-producer/bin/proof-producer/src/main.cpp
@@ -132,7 +132,6 @@ int run_prover(const nil::proof_generator::ProverOptions& prover_options) {
                         prover.fill_assignment_table(prover_options.trace_base_path) &&
                         prover.save_assignment_description(prover_options.assignment_description_file_path) &&
                         prover.preprocess_public_data() &&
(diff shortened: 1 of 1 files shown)


proof-producer/bin/proof-producer/src/main.cpp
```diff
@@ -132,7 +132,6 @@ int run_prover(const nil::proof_generator::ProverOptions& prover_options) {
                         prover.fill_assignment_table(prover_options.trace_base_path) &&
                         prover.save_assignment_description(prover_options.assignment_description_file_path) &&
                         prover.preprocess_public_data() &&
-                        prover.save_preprocessed_common_data_to_file(prover_options.preprocessed_common_data_path) &&
```
akokoshn, line 135: i suspect we may need it as public input. Will send as result together with proof (some how merge for all circuits)

akokoshn (changes requested):

proof-producer/bin/proof-producer/src/main.cpp
makxenov: Thanks, my bad, I've missed it

PR #292, opened 2025-01-22
Ensure that comparisons are not identically false

mizabrik:


crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/scmp.hpp
```diff
@@ -169,8 +176,6 @@ namespace nil {
                     // constrain(diff * (diff * diff_inv - 1));
```
ETatuzova, line 176: Uncomment constraints that are similar to CMP

PR #302, opened 2025-01-28
test(proof producer): add proof generation/verifaction for circuits

x-mass:


x-mass:
Locally it fails due to proto hash missmatch. Waiting untill we have test data generator

proof-producer/tests/bin/proof-producer/test_zkevm_bbf_circuits.cpp
```diff
@@ -106,14 +108,61 @@ TEST(ProverTest, TraceIndexMismatch) {
...
                         ZKEVM
     );
 
     ASSERT_TRUE(prover.setup_prover());
     ASSERT_FALSE(prover.fill_assignment_table(trace_base_path));
 }
+
+namespace fs = std::filesystem;
+
+class ProverTestsVerification : public ProverTestBase {
```
akokoshn, line 125: Suspect these tests takes much time on CI (on local dev server zkevm verification with minimal size takes ~1h).
Check will be duplicated in [regular benchmarks](https://github.com/NilFoundation/placeholder/pull/309), so most probably no need add it to CI

PR #320, opened 2025-02-10
dFRI verifier

martun:


proof-producer/bin/proof-producer/include/nil/proof-generator/commands/verify_aggregated_command.hpp
```diff
@@ -0,0 +1,215 @@
...
+                        constraint_systems.push_back(std::make_shared<ConstraintSystemNoCircuit>(*constraint_systems_[i]));
+                    }
+
+                    bool verification_result = nil::crypto3::zk::snark::placeholder_DFRI_verifier<BlueprintField, PlaceholderParams>::process(
+                            common_datas_,
+                            *agg_proof_,
+                            table_descriptions_,
+                            constraint_systems,
+                            lpc_schemes,
+                            {} // public inputs
```
martun, line 126: Here I skip verification of public inputs matching the hashes in common_data. It was done this way in normal verifier, so I also did this way. Not sure if we should read and verify the public inputs for dFRI.

crypto3/libs/zk/include/nil/crypto3/zk/snark/systems/plonk/placeholder/dFRI_verifier.hpp
```diff
@@ -0,0 +1,279 @@
...
+                                typename fri_type::polynomial_values_type y;
+
+                                if (!nil::crypto3::zk::algorithms::verify_initial_proof_and_return_combined_Q_values<fri_type>(
+                                        agg_proof.aggregated_proof.initial_proofs_per_prover[i].initial_fri_proofs.initial_proofs[query_id], Us[i], poly_maps[i], Vs[i], 
+                                        fri_params, commitments[i], theta, coset_size, domain_size, starting_indexes[i], aggregated_transcripts[i], y, x, x_index 
+                                        )) {
+                                    BOOST_LOG_TRIVIAL(info) << "dFRI Verification failed: initial FRI proof/consistency check verification failed for prover #" << i << ".";
+                                    return false;
+                                }
+                                // for all the provers the values of x and x_index must match, since we're using the same transcript for each prover.
```
martun, line 218: Here I assumed that the values of X must match. That also means that evaluation points must match for all the circuits. For example if some circuit has a lookup argument, and another one has not, we can't work with that circuits.

crypto3/libs/zk/test/systems/plonk/placeholder/placeholder_dFRI_test_runner.hpp
```diff
@@ -0,0 +1,237 @@
...
+              constraint_system1(circuit1_in.gates, circuit1_in.copy_constraints, circuit1_in.lookup_gates, circuit1_in.lookup_tables),
+              constraint_system2(circuit2_in.gates, circuit2_in.copy_constraints, circuit2_in.lookup_gates, circuit2_in.lookup_tables),
+              assignments1(circuit1_in.table),
+              assignments2(circuit2_in.table),
+              table_rows_log(std::log2(circuit1_in.table_rows)),
+              fri_params(1, table_rows_log, placeholder_test_params::lambda, 4) {
+    }
+
+    // The code in this function generates an aggregated dFRI proof using the provided 2 circuits.
+    // It is very similar to the code in the proof-producer. Should something change in the algorithm,
```
martun, line 107: The code of this function should match that in the proof-producer. I believe it matches :)

crypto3/libs/zk/include/nil/crypto3/zk/snark/systems/plonk/placeholder/dFRI_verifier.hpp
ayashunsky: Maybe we should add this info to the error message below, expanding the short " with challenge x mismatch.". Because longer error messages help :)

martun: I think it can be too long, we have it in the comment. We don't want normal users seeing these messages when something fails. The developer will very fast find where the exception comes from, and then will see the comment above. I can add a bit longer comment here.

ayashunsky: Ok, let's settle for a longer comment.

martun: Changed the comment.

ayashunsky (approved):

PR #325, opened 2025-02-11
proof-producer: use proper retcodes

oclaw:
- Added separate OutOfMemory error and mapped it to std::bad_alloc
- Wrapped all IO errors with input/output artifacts operations
- Moved to exceptions in assigner trace parser in order to make a difference between IO errors and malformed or inconsistent traces
- Wrapped all `std::logic_error`  exceptions from crypto3 with ProverError
- Added retcode check to the proof producer tests

### DIFF
--- a/proof-producer/libs/assigner/include/nil/proof-generator/assigner/trace_parser.hpp
+++ b/proof-producer/libs/assigner/include/nil/proof-generator/assigner/trace_parser.hpp
@@ -1,14 +1,15 @@
 #ifndef PROOF_GENERATOR_LIBS_ASSIGNER_TRACE_PARSER_HPP_
 #define PROOF_GENERATOR_LIBS_ASSIGNER_TRACE_PARSER_HPP_
 
+#include <stdexcept>
 #include <utility>
 #include <fstream>
 #include <ios>
 #include <optional>
-#include <vector>
 #include <string>
 #include <cstdint>
-#include <unordered_map>
+#include <format>
+
 #include <boost/filesystem.hpp>
 #include <boost/log/trivial.hpp>
 #include <boost/filesystem.hpp>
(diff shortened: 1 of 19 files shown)


proof-producer/libs/assigner/include/nil/proof-generator/assigner/trace_parser.hpp
```diff
@@ -53,23 +83,34 @@ namespace nil {
...
-            [[nodiscard]] std::optional<ProtoTraces> read_pb_traces_from_file(const boost::filesystem::path& filename) {
+            [[nodiscard]] ProtoTraces read_pb_traces_from_file(
+                const boost::filesystem::path& filename,
+                TraceIndexOpt index_base,
+                const AssignerOptions& options
+            ) {
                 std::ifstream file(filename.c_str(), std::ios::in | std::ios::binary);
                 if (!file.is_open()) {
-                    return std::nullopt;
+                    trace_io_error err(filename.string());
```
akokoshn, line 93: throw?

akokoshn (commented):
Not sure about moving to use exceptions

oclaw:
@akokoshn  Not using exceptions means that we have to:
- introduce specific error codes for assigner
- make every parse function be able to return it
- adjust every assigner function
- adjust assigner function map
- convert these assigner error codes to CommandResult

It requires much more of code structure change and, frankly, I do not see the reason for doing it. Especially since the crypto3 parts are already designed to return an errors via exceptions right now

proof-producer/libs/assigner/include/nil/proof-generator/assigner/trace_parser.hpp
oclaw: yes, thanks

PR #329, opened 2025-02-12
correct the list to match the current implementation status

AndreyMlashkin:
The list of tests for verification is including the elements, that are not passing verification for now


AntoineCyr (approved):

akokoshn (changes requested):
We no need proof-producer proof generation/verification test.
Need to restore EVM verification tests with generation actual inputs and enable only which passed now.

akokoshn (approved):
Ok, let's merge it as WA for unblock CI now

PR #337, opened 2025-02-18
336 crypto3: add mersenne31, koalabear and babybear

ioxid:
Closes #336

PR #339, opened 2025-02-20
Make all BOOST_TEST_MODULE unique in order not to being overwritten

AndreyMlashkin:
This PR is fixes the CI error like this:
https://github.com/NilFoundation/placeholder/pull/337/checks?check_run_id=37423559317

It also should fix sporadic number of tests (right now, test outputs with same names are overridden by latest run)


akokoshn (commented):
Generally looks ok. Let's wait for review from @martun @ioxid

martun (approved):

PR #346, opened 2025-02-25
migrate to new satisfiliability check

oclaw:
Unified old and new implementation of satisfiability check (from the circuit builder and a free function one) and made it multi-threaded

Also deleted assignment and circuit proxy classes from blueprint as there is no need for them anymore


crypto3/libs/blueprint/include/nil/blueprint/blueprint/plonk/circuit.hpp
```diff
@@ -73,6 +73,8 @@ namespace nil {
             gate_selector_map selector_map = {};
             lookup_gate_selector_map lookup_selector_map = {};
             std::size_t next_selector_index = 0;
+            std::map<uint32_t, std::vector<std::string>> constraint_names_;
```
oclaw, line 76: Moved here from circuit_builder

crypto3/libs/blueprint/include/nil/blueprint/blueprint/plonk/assignment_proxy.hpp
```diff
@@ -445,23 +445,6 @@ namespace nil {
             }
             return res;
         }
-
```
akokoshn, line 448: may be it's time to remove both assigner and circuit proxy

crypto3/libs/blueprint/include/nil/blueprint/blueprint/plonk/circuit.hpp
akokoshn: let's move to [zk type](https://github.com/NilFoundation/placeholder/blob/80e139031345d4068c21c1fd60e1143a99c00162/crypto3/libs/zk/include/nil/crypto3/zk/snark/arithmetization/plonk/constraint_system.hpp#L58)

oclaw: Not sure about it. It is (more or less) a field with meta information about the circuit, don't know if the core type is a good place for it. @mizabrik what do you think?

akokoshn: @oclaw , another way - pass names as optional parameter of `is_satisfied`, and return `map` back to `circuit_builder`.
Will be used, when call `circuit_builder::is_satisfied()`

oclaw: It is a bad idea. Circuit builder is an internal BBF utility which should not be exposed outside of preset/assigner lib. In this case we will have to adjust signatures of all functions - circuit initializers and pollute the code with debug only fields.

The good decision here would be stop using blueprint::circuit in proof-producer and reduce it to plonk_constraint_system type inside the preset command step (so all debugging stuff can be released in case if no one needs it), but it is a topic for another PR, should not be done here

oclaw: Or you mean make it part of the sat check options and skip it while checking externally? It can be the way, yes (but will reduce readability of the proof-producer tests)

oclaw: Yes, it is the best solution I think, thanks!
Done

crypto3/libs/blueprint/include/nil/blueprint/bbf/circuit_builder.hpp
```diff
@@ -728,7 +512,7 @@ namespace nil {
 
                     circuit<crypto3::zk::snark::plonk_constraint_system<FieldType>> bp;
                     crypto3::zk::snark::plonk_assignment_table<FieldType> presets;
-                    std::map<uint32_t, std::vector<std::string>> constraint_names;
+                    std::map<uint32_t, std::vector<std::string>> constraint_names_;
```
akokoshn, line 515: why not just `constraint_names`, without `_`?

akokoshn (approved):

crypto3/libs/blueprint/include/nil/blueprint/bbf/circuit_builder.hpp
oclaw: fixed (but in general I prefer to keep difference between class fields and local variables)

akokoshn (approved):

PR #365, opened 2025-02-28
354 add comments to not opcode

Valeh2012:


crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/not.hpp
```diff
@@ -50,35 +63,41 @@ namespace nil {
...
+                        // split a (stack top) to 16-bit chunks
                         auto a = w_to_16(current_state.stack_top());
                         for( std::size_t i = 0; i < 16; i++ ){
                             A[i] = a[i];
-                            R[i] = 0xFFFF - a[i];
                         }
                     }
+                    
+                    /* Layout:         range_checked_opcode_area 
+                        +------+------+------+------+------+------+------+------+
```
ayashunsky, line 79: We don't need the W's, but column numbers could still be useful especially in more complex circuits.

PR #380, opened 2025-03-13
Fix "Illegal instruction" crash on MacOS in tests

mizabrik:


AntoineCyr:
It fixes the issue for me as well :)

AntoineCyr (approved):

PR #406, opened 2025-04-03
405 sar tests

AntoineCyr:
3 changes:

- SCMP now correctly act when a and b are both negatives.
- SAR is now SHR + signextend, where signextend is slightly different than the opcode because SAR is more precise (256 possible values instead of 32)
- Modify a constraint for call_commit, the constraint was failing when retracing a pure function because there is no items.


ayashunsky:
Since we're redoing the SAR anyway, I'd opt for more comments and better variable names.

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/sar.hpp
```diff
@@ -253,108 +258,105 @@ namespace nil {
...
+                        // when shift is 256, n = 0 => no transition chunk
+                        
+                        zkevm_word_type rn_temp = (n == 0) ? 0 : field_to_int(r_chunks[field_to_int(n - 1)]); 
+                        // (n-1)-th chunk of r, for further computations only
+                        full_shift = (n==0);
+
+                        std::cout << "b0: " << b0 << std::endl;
+                        std::cout << "n: " << n << std::endl;
+                        std::cout << "parity: " << parity << std::endl;
+
```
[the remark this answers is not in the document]
AntoineCyr: You are right, I got rid of indic_3 and indic_4

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/sar.hpp
```diff
@@ -134,423 +144,434 @@ namespace nil {
...
+                    constrain(input_b_chunks[0] - shift_lower - 16 * shift_middle -
+                              256 * shift_upper);
+                    constrain(shift_upper * (1 - shift_upper * upper_inverse));
+
+                    TYPE partial_sum_check = 1 - b_partial_sum * sum_inverse_partial;
+                    allocate(partial_sum_check, 43, 1);
+                    constrain(b_partial_sum * partial_sum_check);
+                    constrain(is_shift_large -
+                              (1 - shift_upper * upper_inverse) * partial_sum_check);
+
```
ayashunsky, line 403: Many of the above constraints can be defined automatically by providing a variable definition outside the assignment stage. For instance if you place "b_is_zero = 1 - b_sum_inverse * b_total_sum;" outside the assignment stage block, allocation of b_is_zero will automatically generate the constraint "1 - b_sum_inverse * b_total_sum - b_is_zero = 0". It seems that current more-or-less best practice is to have a sequence of multiple blocks of "Assignment/definition -> allocation -> Contraints (this one is actually optional)", one following the other. This allows to make best use of implicit constraints for the cases like the one I wrote about above.

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/sar.hpp
```diff
@@ -134,423 +144,434 @@ namespace nil {
...
+                        // Decompose shift amount
+                        shift_lower = input_b % 16;
+                        shift_middle = (input_b / 16) % 16;
+                        shift_upper = (input_b % 65536) / 256;
+                        upper_inverse =
+                            shift_upper.is_zero() ? 0 : shift_upper.inversed();
+
+                        // Calculate sums and inverses
+                        b_partial_sum = 0;
+                        b_total_sum = std::accumulate(b_chunks.begin(), b_chunks.end(),
```
ayashunsky, line 249: we can do 
b_total_sum = b_partial_sum + b_chunks[0]
after the cycle below.

AntoineCyr: I don't see how, b_partial_sum is constructed by input_b_chunks, and b_total_sum by b_chunks

ayashunsky: ok, sorry, I missed the fact that they were sums of different b's

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/sar.hpp
AntoineCyr: I wrote b_is_zero and is_shift_large like this, I don't think I see other way to use implicit constraints.

ayashunsky (approved):

PR #436, opened 2025-04-29
Optimizing the dag expression by removing duplicate nodes.

martun:


parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
```diff
@@ -239,13 +288,320 @@ namespace nil::crypto3::zk::snark {
...
+                    }
+                }
+            }
+            std::vector<std::pair<size_t, size_t>> selected_pairs;
+            std::vector<bool> used(result.nodes.size(), false);
+            for (const auto& [key, pr] : pairs) {
+                const auto& [first, second] = pr;
+                if (used[first] || used[second])
+                    continue;
+                used[first] = used[second] = true;
```
ioxid, line 419: Better have this on two separate lines, this chaining is a bit unreadable

ioxid (commented):
Do we have consensus on std::size_t vs plain size_t? I'm accustomed to std::size_t because size_t is in global namespace only as a C leftover, C++ standard says it's in std. Anyway, reading with this std:: removal is a bit hard, better have it in a separate commit.

martun:
> Do we have consensus on std::size_t vs plain size_t? I'm accustomed to std::size_t because size_t is in global namespace only as a C leftover, C++ standard says it's in std. Anyway, reading with this std:: removal is a bit hard, better have it in a separate commit.

I like size_t better, since it's shorter. I do understand that std::size_t is in std namespace, so it's newer or more correct, but anyway...

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
```diff
@@ -325,9 +329,14 @@ namespace nil::crypto3::zk::snark {
...
+
             for (const auto& expr: expressions) {
                 size_t root_node = boost::apply_visitor(*this, expr.get_expr());
                 result.root_nodes.push_back(root_node);
+
+                std::size_t degree = max_degree_visitor.compute_max_degree(expr);
+                result.root_node_degrees.push_back(degree);
             }
             squash_dag();
             return std::move(result);
```
ioxid, line 342: This `move` is wrong

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression_evaluator.hpp
```diff
@@ -93,8 +93,14 @@ namespace nil::crypto3::zk::snark {
             const size_t extended_domain_size = _cached_assignment_table.original_domain_size * _max_degree;
 
             // Create empty dfs polynomials of degree 'extended_domain_size - 1' and size 'extended_domain_size'.
-            std::vector<polynomial_dfs_type> result(_expr.get_root_nodes_count(),
-                polynomial_dfs_type(extended_domain_size - 1, extended_domain_size));
```
ioxid, line 90: This already used extended_domain_size, so what's the difference? It is different for different roots now?

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression_evaluator.hpp
```diff
@@ -153,6 +200,8 @@ namespace nil::crypto3::zk::snark {
             }
         }
 
+        std::vector<size_t> _occurences;
+        std::vector<bool> _is_root;
```
ioxid, line 204: `vector<bool>` is very slow, use `vector<char>`

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression_evaluator.hpp
```diff
@@ -136,15 +151,47 @@ namespace nil::crypto3::zk::snark {
...
                     const auto& add = std::get<dag_addition>(node);
-                    assignment_chunks[k] = assignment_chunks[add.operands[0]];
-                    for (std::size_t i = 1; i < add.operands.size(); i++) {
-                        assignment_chunks[k] += assignment_chunks[add.operands[i]];
+                    // Find a child we can move.
+                    bool moved_a_child = false; 
+                    for (std::size_t i = 0; i < add.operands.size(); i++) {
+                        if ((_occurences[add.operands[i]] == 1) && !_is_root[add.operands[i]]) {
+                            moved_a_child = true; 
+                            assignment_chunks[k] = std::move(assignment_chunks[add.operands[i]]);
```
ioxid, line 159: This `move` does nothing right now, because `static_simd_vector` is stored inplace. `assignment_chunks` should store a `unique_ptr` to have indirection, or some other way.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
```diff
@@ -212,10 +218,94 @@ namespace nil::crypto3::zk::snark {
...
+        std::vector<size_t> _occurences;
+
+        dag_child_occurence_counting_visitor() = default;
+
+        std::vector<size_t> get_occurence_counts(const dag_expression<VariableType> &dag) {
+            _occurences = std::vector<size_t>(dag.get_nodes_count(), 0);
+            for (const auto& node: dag.get_nodes()) {
+                std::visit(*this, node);
+            }
+            return std::move(_occurences);
```
ioxid, line 281: This `move` is wrong

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
```diff
@@ -238,14 +329,351 @@ namespace nil::crypto3::zk::snark {
...
+            //std::cout << "DAG After remove unreachable nodes" << std::endl;
+            //visitor.print_stats(result);
+
+            remove_duplicates();
+
+            //std::cout << "DAG After removing duplicates" << std::endl;
+            //visitor.print_stats(result);
+
+            // Now remove children that are the only have one parent, and parent operation matches.
+            merge_children(true);
```
ioxid, line 370: Add what parameter is `true`, like `/*only_those_that_occur_once=*/ true`. Then maybe the comment on previous line is not really necessary.

ioxid (changes requested):

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
martun: Why is it wrong. It moves out a class member, knowing that it's not required any more.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression_evaluator.hpp
martun: Now I put a correct degree for each result. We will have constraints of different degrees, some for example will have degree 3, I.E. the resulting polynomial DFS will have degree 3*N, while the size will be 4*N. Previously we were setting the degree to 4*N for all.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression_evaluator.hpp
martun: Let's keep it, it's more readable, and we don't care about performance here.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
martun: Why do you think so, I'm moving out a member variable. If I remove it, it will copy.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
martun: Done.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression_evaluator.hpp
martun: You're right. Initially I changed std::array to std::vector in simd_vector_type, but then decided to change it back. Let's use unique_ptr here.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression_evaluator.hpp
martun: Removed this code.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression_evaluator.hpp
martun: I changed it to unique_ptr, it slowed down. Probably since we move it, we need to allocate it again on the next iteration, which is slower than copying the data inside. Rolled back this part of the change.

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
ioxid: Oh, if it's a class member then it's okay. I though it's a local variable

parallel-crypto3/libs/parallel-zk/include/nil/crypto3/zk/math/dag_expression.hpp
ioxid: I once again though it's a local variable. This time it even has an underscore...

PR #439, opened 2025-05-07
438 use fft in batches

martun:


parallel-crypto3/libs/parallel-math/include/nil/crypto3/math/polynomial/polynomial_dfs.hpp
```diff
@@ -919,8 +924,10 @@ namespace nil {
             }
 
             template<typename FieldType>
-            static inline polynomial_dfs<typename FieldType::value_type> polynomial_product(
```
ioxid, line 922: Why not static?

parallel-crypto3/libs/parallel-math/include/nil/crypto3/math/domains/basic_radix2_domain.hpp
```diff
@@ -91,13 +93,60 @@ namespace nil {
...
+                                p.resize(this->m, value_type::zero());
+                            } else {
+                                throw std::invalid_argument("Expected polynomail size <= domain size");
+                            }
+                        }
+                    }
+                }
+
+                /** \brief Batch version of the 'fft' function. Used for vectorization.
+                 *  \param[in] a - Each element of a a[i] represents coefficients of a polynomial.
```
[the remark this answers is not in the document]
martun: Changed the comment.

parallel-crypto3/libs/parallel-math/include/nil/crypto3/math/polynomial/polynomial_dfs.hpp
martun: It makes no sense, this is not a member function. Adding static changes nothing, we are header only, it's just 1 compilation unit anyway. Inline is probably useless as well, I don't believe compiler will decide to inline this.

parallel-crypto3/libs/parallel-math/include/nil/crypto3/math/domains/basic_radix2_domain.hpp
```diff
@@ -91,13 +93,60 @@ namespace nil {
...
-                            throw std::invalid_argument("basic_radix2: expected a.size() == this->m");
+                            throw std::invalid_argument("basic_radix2: expected a.size() <= this->m");
                         }
                     }
 
                     detail::basic_radix2_fft_cached<FieldType>(a, fft_cache->first);
                 }
 
+                void resize_to_domain_size(std::vector<std::vector<value_type>> &a) {
+                    PARALLEL_PROFILE_SCOPE("resize_to_domain_size {} vectors from size {} to {}",
```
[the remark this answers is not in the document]
martun: Yes, changed to PROFILE_SCOPE.

parallel-crypto3/libs/parallel-math/include/nil/crypto3/math/polynomial/polynomial_dfs.hpp
ioxid: Oh, I thought it's inside a class. Yes, we need to remove all statics from global functions in headers.
Inline is not about inlining and actually is about One Definition Rule. This function should be inline, like all global functions in header files, but writing it explicitly is unnecessary because all templates are implicitly inline.
So yeah, removing static and inline is right.

parallel-crypto3/libs/parallel-math/include/nil/crypto3/math/polynomial/polynomial.hpp
```diff
@@ -142,7 +142,7 @@ namespace nil {
                 }
 
                 polynomial& operator=(polynomial&& x) {
```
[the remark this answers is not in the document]
ioxid: I would assume not, because the other one is not move assignment, it's just assignment from unrelated type (which happens to be rvalue reference).

parallel-crypto3/libs/parallel-math/include/nil/crypto3/math/polynomial/polynomial.hpp
```diff
@@ -136,15 +132,9 @@ namespace nil {
...
-                    val = x.val;
-                    return *this;
-                }
+                polynomial& operator=(const polynomial& x) = default;
 
-                polynomial& operator=(polynomial&& x) {
-                    val = std::move(x.val);
-                    return *this;
-                }
+                polynomial& operator=(polynomial&& x) = default;
```
ioxid, line 137: Can you try removing all these defaulted constructors/operator=?

PR #441, opened 2025-05-12
SHR opcode circuit adapted for small fields

javiersilva0:
Basically the same changes as in the SAR circuit, except for the sign extension part.

In the case of a large shift, we simply set 0 as the result.

### DIFF
--- a/crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/shr.hpp
+++ b/crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/shr.hpp
@@ -30,437 +30,441 @@
 #include <nil/blueprint/zkevm_bbf/types/opcode.hpp>
 #include <numeric>
 
-namespace nil {
-    namespace blueprint {
-        namespace bbf {
-            template<typename FieldType>
-            class opcode_abstract;
+namespace nil::blueprint::bbf {
+template<typename FieldType>
+class opcode_abstract;
 
-            template<typename FieldType, GenerationStage stage>
-            class zkevm_shr_bbf : public generic_component<FieldType, stage> {
-                using generic_component<FieldType, stage>::allocate;
-                using generic_component<FieldType, stage>::copy_constrain;
-                using generic_component<FieldType, stage>::constrain;
-                using generic_component<FieldType, stage>::lookup;
-                using generic_component<FieldType, stage>::lookup_table;
+template<typename FieldType, GenerationStage stage>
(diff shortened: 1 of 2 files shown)


crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/shr.hpp
```diff
@@ -30,414 +31,441 @@
...
+    constexpr static const std::size_t chunk_8_amount = 32;
+    constexpr static const value_type two_15 = 32768;
+    constexpr static const value_type two_16 = 65536;
+
+    public:
+    using typename generic_component<FieldType, stage>::TYPE;
+    using typename generic_component<FieldType, stage>::context_type;
+
+    // Computes the terms of r*b with coefficient 2^(16 * chunk_index)
+    template<typename T, typename V = T>
```
ayashunsky, line 59: This line is one-year-old legacy. You can most probably forget about V now, drop the "template" line and replace T with TYPE. Both here and in the function below.

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/opcodes/shr.hpp
```diff
@@ -30,414 +31,441 @@
...
+
+        // PART 1: computing the opcode and splitting values in chunks
+        if constexpr (stage == GenerationStage::ASSIGNMENT) {
+            // Extract input values from stack
+            zkevm_word_type input_b = current_state.stack_top();    // Shift amount
+            zkevm_word_type a = current_state.stack_top(1);         // Value to shift
+
+            // Calculate shift and result
+            // any shift of 256 bits or more is equivalent
+            // to a 255 (NB! not 256) bit shift, which leaves us with just the sign bit
```
ayashunsky, line 146: This comment and the logic that follows was only valid for SAR. The SHR opcode should shift 256 bits when asked to shift 256. This requires changes in the logic below.

javiersilva0: The comment is misleading, I updated it with a high-level summary of what I tried to accomplish. I'm considering a 256-bit shift a large one. And in this case, even if the shift logic computes a 255-bit shift, the result is set to 0 anyway. See lines 155 and 399 for assignment and constrain, respectively.

Let me know if you want me to expand the comment further.

ayashunsky (approved):

PR #472, opened 2025-06-02
codecopy

### LINKED ISSUE #455: Bytecode circut for small fields underconstrained
- [ ] Constrain CBOR-metadata length and placement at the end of the bytecode
- [ ] Implement CODECOPY, EXTCODECOPY opcode. Test it with copied bytecode data piece contain metadata and handle overflow

AntoineCyr:


ETatuzova:
Uncomment this row to check lookup from copy circuit to bytecode table:

https://github.com/NilFoundation/placeholder/blob/455-bytecode-small-field/crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/small_field/circuits/copy.hpp#L406

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/small_field/subcomponents/memory_cost.hpp
```diff
@@ -0,0 +1,79 @@
...
+                integral_type memory2 =
+                    integral_type(mem_words.to_integral()) * integral_type(mem_words.to_integral());
+                mem_cost = memory2 / 512 + 3 * integral_type(mem_words.to_integral());
+                R = memory2 % 512;
+            }
+
+            allocate(mem_words, 3, 0);
+            allocate(mem_cost, 4, 0);
+            allocate(R, 5, 0);
+            constrain((mem_cost - 3 * mem_words) * 512 - mem_words * mem_words +
```
ETatuzova, line 72: Can we assume that `mem_words * mem_words` fits into field size?

ETatuzova: I tried to estimate maximum memory size we have to support.

If we have 2^21 bound then max `mem_words` will be 2^16, then memory cost for this will be
`(3*(2^32))/512 + 3 * 2^16 = 25 362 432`. It is smaller than 36M.
So, it's not enough.

Let's try 2^22. Max `mem_words = 2^17` 
`(3*(2^34))/512 + 3 * 2^7= 101056512`.
It doesn't fit into 36M, but fits into small field. 
It seems a best option.

But 2^34 doesn't fit into 2^31. So, circuit should be more complex.

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/input_generators/basic_evm.hpp
```diff
@@ -1526,21 +1526,38 @@ namespace nil {
...
+                    // max contract size is 24,576 bytes, so offset and length need to fit in
+                    // first chunk
+                    constexpr static const std::size_t max_offset = 65536;
+                    constexpr static const std::size_t max_length = 65536;
                     std::size_t dst = std::size_t(stack.back()); stack.pop_back();
                     std::size_t src = std::size_t(stack.back()); stack.pop_back();
                     std::size_t length = std::size_t(stack.back()); stack.pop_back();
 
+                    bool overflow = (dst > max_dest_offset) ||
+                                    (src > max_offset) ||
```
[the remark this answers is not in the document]
AntoineCyr: You are right, the offset value should never create an error.

crypto3/libs/blueprint/include/nil/blueprint/zkevm_bbf/input_generators/basic_input_generator.hpp
```diff
@@ -1000,29 +1000,45 @@ namespace nil {
...
+                        copy_event cpy = codecopy_copy_event(
+                            bytecode_hash,
+                            source,
+                            call_id,
+                            distance,
+                            rw_counter,
+                            real_length
+                        );
+                        //TODO?
+                        //Do we need a second copy event for the fake copy?
```
ETatuzova, line 1028: 1. Add this copy event only if real_length != 0.
2. Do we check EVM's behaviour? If 0 is written to the remaining memory, an additional copy event is needed

ETatuzova:
Merged with gas_error PR

PR #477, opened 2025-06-06
Clean the code of LPC, use addition point _etha much easier. Don't convert committed polynomials to coefficients form twice. Allow committing to polynomials only in DFS form in LPC.

martun:
- Clean the code of LPC
- Use addition point _etha much easier
- Don't convert committed polynomials to coefficients form twice.
- Allow committing to polynomials only in DFS form in LPC.

### DIFF
--- a/crypto3/libs/zk/include/nil/crypto3/zk/commitments/polynomial/lpc.hpp
+++ b/crypto3/libs/zk/include/nil/crypto3/zk/commitments/polynomial/lpc.hpp
@@ -50,12 +50,11 @@ namespace nil {
         namespace zk {
             namespace commitments {
 
-                // Placeholder-friendly class.
                 // LPCScheme is usually 'batched_list_polynomial_commitment<...>'.
-                template<typename LPCScheme, typename PolynomialType = typename math::polynomial_dfs<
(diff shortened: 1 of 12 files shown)


crypto3/libs/zk/include/nil/crypto3/zk/commitments/polynomial/lpc.hpp
```diff
@@ -294,91 +324,45 @@ namespace nil {
...
-                        // it will point to this->_polys if no conversion is required, or to polys_coefficients
-                        // if conversion was required.
-                        std::map<std::size_t,
-                                 std::vector<typename PolynomialType::polynomial_type>>
-                            polys_coefficients;
-                        std::map<std::size_t,
-                                 std::vector<typename PolynomialType::polynomial_type>>*
-                            polys_coefficients_ptr;
-
-                        if constexpr (std::is_same<math::polynomial_dfs<value_type>,
```
ioxid, line 348: Are you sure we wont need this? Is there any chance that changing representation will make things different (e.g. less FFTs)?

ioxid (approved):

PR #484, opened 2025-06-18
feat: efficient eval at point in FRI

ioxid:


crypto3/libs/zk/include/nil/crypto3/zk/commitments/polynomial/lpc.hpp
```diff
@@ -328,7 +328,6 @@ namespace nil {
                         PROFILE_SCOPE("LPC prepare combined Q");
 
                         this->build_points_map();
-                        convert_polys_to_coefficients_form();
```
martun, line 331: Probably if you remove it here, dFRI tests will fail, because you need to add this call there as well.

martun (commented):
Added 2 comments

crypto3/libs/zk/include/nil/crypto3/zk/commitments/polynomial/lpc.hpp
ioxid: I added the required call to `prover.hpp`

